PR #1, opened 2025-08-28
[sandrej09] ip

sandrej09:


src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
...
+        System.out.println(LINE);
+        System.out.println(" Hello! I'm Lumi");
+        System.out.println(" What can I do for you?");
+        System.out.println(LINE);
+
+        while (true) {
+            String input = sc.nextLine().trim();
+
+            if (input.equals("bye")) {
+                System.out.println(LINE);
```
HarshitSrivastavaHS: Can add this as a function printLine() and just call it instead of adding print statement everytime

src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
...
+                System.out.println(" OK, I've marked this task as not done yet:");
+                System.out.println("   " + tasks[idx]);
+                System.out.println(LINE);
+            } else {
+                tasks[count] = new Task(input);
+                count++;
+                System.out.println(LINE);
+                System.out.println(" added: " + input);
+                System.out.println(LINE);
```
HarshitSrivastavaHS: can also try switch case instead of if statement (this is totally fine as well)

src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
...
+                }
+                System.out.println(LINE);
+            } else if (input.startsWith("mark ")) {
+                int idx = Integer.parseInt(input.substring(5)) - 1;
+                tasks[idx].mark();
+                System.out.println(LINE);
+                System.out.println(" Nice! I've marked this task as done:");
+                System.out.println("   " + tasks[idx]);
+                System.out.println(LINE);
+            } else if (input.startsWith("unmark ")) {
```
HarshitSrivastavaHS: can try to store a substring in a variable and just match the content of the variable with the command instead of checking if it starts with it. (like store the substring from index 0 to the first instance of space)

src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
...
+        System.out.println(LINE);
+
+        while (true) {
+            String input = sc.nextLine().trim();
+
+            if (input.equals("bye")) {
+                System.out.println(LINE);
+                System.out.println(" Bye. Hope to see you again soon!");
+                System.out.println(LINE);
+                break;
```
HarshitSrivastavaHS: this works, however can use System.exit(0) to terminate the java program

src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
...
+                int idx = Integer.parseInt(input.substring(7)) - 1;
+                tasks[idx].unmark();
+                System.out.println(LINE);
+                System.out.println(" OK, I've marked this task as not done yet:");
+                System.out.println("   " + tasks[idx]);
+                System.out.println(LINE);
+            } else {
+                tasks[count] = new Task(input);
+                count++;
+                System.out.println(LINE);
```
HarshitSrivastavaHS: since you're printing this line in every if block, can just print it before entering the if block and at the end after exiting

HarshitSrivastavaHS (commented):
Looks pretty good though :))

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
+public class Task {
```
limzerui: I noticed there’s no package declaration here. According to the coding standard, every class should be in a lower-case package name (e.g., lumi.task )

src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
+import java.util.Scanner;
+
+public class Lumi {
+    private static final String LINE = "____________________________________________________________";
+
+    public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
+        Task[] tasks = new Task[100];
```
limzerui: nit: The 100 here feels like a magic number. Would it be clearer to either (a) define a named constant like MAX_TASKS, or (b) use a dynamic collection (e.g., List<Task>) to avoid the fixed limit and extra count bookkeeping?

src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
...
+                break;
+            } else if (input.equals("list")) {
+                System.out.println(LINE);
+                System.out.println(" Here are the tasks in your list:");
+                for (int i = 0; i < count; i++) {
+                    System.out.println(" " + (i + 1) + ". " + tasks[i]);
+                }
+                System.out.println(LINE);
+            } else if (input.startsWith("mark ")) {
+                int idx = Integer.parseInt(input.substring(5)) - 1;
```
limzerui: I like that you’re handling mark and unmark here 👍. One improvement could be to wrap the parsing and indexing in a try/catch.
Right now, invalid input (like mark abc or mark 200) would crash with NumberFormatException or ArrayIndexOutOfBoundsException.

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
...
+    private final String description;
+    private boolean isDone;
```
limzerui: i like that the boolean field is named isDone with an is prefix. this follows the convention for boolean varialbels and makes conditions like if read naturally

src/main/java/Task.java
limzerui: Good job using PascalCase for the class name which matches the stadnard that class names must be nouns in PascalCase

limzerui (commented):
code is generally clean and follows several Java coding standards. 
have provided a few suggestions such as not using magic numbers, exception handling and package declaration

src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
+import java.util.Scanner;
+
+public class Lumi {
+    private static final String LINE = "____________________________________________________________";
+
+    public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
+        Task[] tasks = new Task[100];
+        int count = 0;
```
muadzyamani: The variable name `count` works, but could we make it more descriptive? Something like `taskCount` or `numberOfTasks` might make the code more self-documenting. What do you think?

src/main/java/Lumi.java
```diff
@@ -0,0 +1,54 @@
...
+
+        System.out.println(LINE);
+        System.out.println(" Hello! I'm Lumi");
+        System.out.println(" What can I do for you?");
+        System.out.println(LINE);
+
+        while (true) {
+            String input = sc.nextLine().trim();
+
+            if (input.equals("bye")) {
```
muadzyamani: Nice use of `.trim()` to handle whitespace 👍 Just curious, should the commands be case-sensitive? Would it be user-friendly to handle "BYE" or "Bye" as well?

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public void unmark() {
+        isDone = false;
+    }
+
+    @Override
+    public String toString() {
+        String status = isDone ? "[X]" : "[ ]";
+        return status + " " + description;
+    }
```
muadzyamani: Great use of the ternary operator here. The formatting is clean and readable. I like how you've overridden `toString()` to provide a meaningful representation.

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
...
+    private boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public void mark() {
+        isDone = true;
+    }
```
muadzyamani: These methods are nice and concise. I appreciate how the method names clearly indicate what they do.

muadzyamani (commented):
I like how clean and straightforward this implementation is so far. The basic functionality is solid and the code structure is easy to follow. The separation between the main application logic in `Lumi.java` and the task representation in `Task.java` shows good understanding of basic object-oriented principles. Looking forward, perhaps you can think if the main method could benefit from being broken into smaller, focused class and methods.

src/main/java/lumi/Lumi.java
```diff
@@ -0,0 +1,174 @@
+package lumi;
+
+import java.io.*;
+import java.nio.file.*;
```
yentheng0110: Imported classes should be listed explicitly (a coding standard followed in CS2113 course)

src/main/java/lumi/Lumi.java
```diff
@@ -0,0 +1,174 @@
...
+    private static final String TAG_BY = "/by";
+    private static final String TAG_FROM = "/from";
+    private static final String TAG_TO = "/to";
+
+    private static final Path DATA_DIR = Paths.get("data");
+    private static final Path DATA_FILE = DATA_DIR.resolve("duke.txt");
+
+    public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
+        Task[] tasks = new Task[100];
```
yentheng0110: Perhaps you can consider defining a constant at the class level for the magic number 100 here

src/main/java/lumi/Lumi.java
```diff
@@ -0,0 +1,174 @@
...
+        int count = load(tasks);
+
+        System.out.println("Hello! I'm Lumi");
+        System.out.println("What can I do for you?");
+
+        while (true) {
+            if (!sc.hasNextLine()) break;
+            String input = sc.nextLine().trim();
+            if (input.isEmpty()) continue;
```
yentheng0110: The coding standard is to have the if statement being in the next line after if, this applies to all parts of your code too

src/main/java/lumi/Task.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public void mark() { done = true; }
+
+    public void unmark() { done = false; }
+
+    protected String status() { return done ? "X" : " "; }
+
+    public boolean isDone() { return done; }
+
+    public String getDescription() { return description; }
```
yentheng0110: The methods should not be written in a single line - a coding standard followed in CS2113

src/main/java/lumi/Todo.java
```diff
@@ -0,0 +1,9 @@
+package lumi;
+
+public class Todo extends Task {
+    public Todo(String description) { super(description); }
+
+
```
yentheng0110: Perhaps you can consider remove 1 blank line (there are 2 blank lines here) for improved readability

yentheng0110 (commented):
Overall, CS2113 coding standards should be followed when you code, changes were suggested

PR #9, opened 2025-08-30
[Samika] iP

samika2005:


src/main/java/Anna.java
```diff
@@ -0,0 +1,89 @@
...
+    Scanner in = new Scanner(System.in);
+
+    String line = "____________________________________________________________";
+    int len = line.length();
+    System.out.println(line);
+    System.out.println("Hello! I'm Anna!");
+    System.out.println("What can i do for you?");
+    System.out.println(line);
+    int c = 0;
+    Task[] list = new Task[100];
```
m-xrtin: Good! Might want to add something about the constant (Like we learnt in the lecture)

src/main/java/Anna.java
```diff
@@ -0,0 +1,89 @@
...
+    for (int i = 1; i <= 100; i++) {
+    String word = in.nextLine();
+    c = c + 5;
+    list[i] = new Task(word);
+    if(word.equals("bye")) {
+    break;
+    }
+    else {
```
m-xrtin: Might wanna do case instead of else if to simplify!

src/main/java/Anna.java
```diff
@@ -0,0 +1,89 @@
+import java.util.*;
+import java.lang.*;
+public class Anna {
```
m-xrtin, line 21: Nicely done for the coding standards with indents

src/main/java/Anna.java
```diff
@@ -0,0 +1,89 @@
...
+    public static void main(String[] args) {
+    Scanner in = new Scanner(System.in);
+
+    String line = "____________________________________________________________";
```
m-xrtin: I think you might want to break up the tasks into different classes instead of all in one file, but the code is looking good!

tam308: line here should be defined as a static variable, and defined in SCREAMING_SNAKE_CASE

src/main/java/Anna.java
tam308: Task[100] here should be avoided as it is a magic number. consider using a constant instead

src/main/java/Anna.java
```diff
@@ -0,0 +1,89 @@
...
+    System.out.println("Hello! I'm Anna!");
+    System.out.println("What can i do for you?");
+    System.out.println(line);
+    int c = 0;
+    Task[] list = new Task[100];
+    for (int i = 1; i <= 100; i++) {
+    String word = in.nextLine();
+    c = c + 5;
+    list[i] = new Task(word);
+    if(word.equals("bye")) {
```
tam308: consider using a switch case to simplify code!

src/main/java/Anna.java
```diff
@@ -0,0 +1,89 @@
...
+            System.out.println("Got it. I've added this task:");
+            String description = word.substring(9,word.indexOf('/'));
+            String by = word.substring(word.indexOf('/') + 3);
+            list[i] = new Deadline(description,by);
+            System.out.println(list[i].toString());
+            System.out.println("Now you have " + i + " tasks in the list.");
+        }
+        else if(word.startsWith("event")) {
+            System.out.println("Got it. I've added this task:");
+            String description = word.substring(6,word.indexOf('/'));
```
tam308: avoid using magic numbers like 6, consider abstracting into a static string

src/main/java/Anna.java
```diff
@@ -0,0 +1,89 @@
...
+    public static void main(String[] args) {
+    Scanner in = new Scanner(System.in);
+
+    String line = "____________________________________________________________";
+    int len = line.length();
+    System.out.println(line);
+    System.out.println("Hello! I'm Anna!");
+    System.out.println("What can i do for you?");
+    System.out.println(line);
+    int c = 0;
```
tam308: c here is ambiguous, consider renaming it

src/main/java/Anna.java
```diff
@@ -0,0 +1,133 @@
+import duke.*;
```
Charly2312, line 2: try not to import using * as it violates the coding standard since imports must be listed explicitly

src/main/java/Anna.java
```diff
@@ -0,0 +1,133 @@
...
+       try {
+           if (word.equals("list")) {
+               System.out.println("Here are the tasks in your list:");
+               for (int j = 1; j < i; j++) {
+                   System.out.println(j + ". " + list[j].toString());
+               }
+               i = i - 1;
+               continue;
+           } else if (word.startsWith("mark")) {
+               Scanner skip = new Scanner(word);
```
Charly2312: you might want to consider abstracting some parts of the function into smaller functions like handleMarkCommadn(), handleTodoCommand(). This will make your code look neater and more readable

src/main/java/Anna.java
```diff
@@ -0,0 +1,133 @@
...
+               if (!skip.hasNext()) {
+                   throw new DukeException("Finish the sentence!!");
+               }
+               skip.next();
+               if (!skip.hasNextInt()) {
+                   throw new DukeException("You know of numbers don't you? Use them.");
+               }
+               int index = skip.nextInt();
+               if (index < 1 || index >= i || list[index] == null){
+                   throw new DukeException("Did you even come up with that many tasks in the first place?" + index);
```
Charly2312: for these kind of statements, you can make them into a constant static string like String INVALID_TASKS = "did you even..."

src/main/java/Anna.java
```diff
@@ -0,0 +1,133 @@
...
+           System.out.println(" " + e.getMessage());
+           System.out.println(line);
+       }
+        System.out.println(line.substring(0,c) + "(._.)" + line.substring(c+5,len));
+    }
+    }
+    System.out.println(line);
+    System.out.println("Bye. Hope to see you again soon!");
+    System.out.println(line);
+    }
```
Charly2312, line 197: try to make it a good habit that functions must be as concise as possible. For now, try to make each function at most 30 lines long

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,24 @@
+package duke;
+
+public class Task {
+    protected String word;
+    protected boolean flag;
```
Charly2312: boolean variables should be named like a question so something like "isFlag"

Charly2312 (commented):
overall, good readability and coding standard. just take note of the minor adjustments that you need to make to improve your code's readability

PR #13, opened 2025-08-30
[HE Pui San] iP

Kennahh:


src/main/java/Chauncey.java
```diff
@@ -0,0 +1,123 @@
...
+            tasks[numOfTask] = new Deadline(deadlineDetails[0].trim(), deadline);
+            break;
+        case "event":
+            String[] eventDetails = task.split("/");
+            String startTime = eventDetails[eventDetails.length-2].trim();
+            String endTime = eventDetails[eventDetails.length-1].trim();
+            tasks[numOfTask] = new Event(eventDetails[0].trim(), startTime, endTime);
+            break;
+        default:
+            System.out.println("Invalid task type.");
```
kevinlokewy: This part of the code is good, because I can understand what you're doing. Also, the indentation for the case switches are correct!

src/main/java/Chauncey.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+        numOfTask++;
+        System.out.println("Got it. I've added this task: ");
+        tasks[numOfTask-1].outputTaskDetails();
+        System.out.println("Now you have " + numOfTask + " tasks in the list.");
+    }
+
+    private static void removeTask(String command) {
+        int taskNumber = Character.getNumericValue(command.charAt(command.length()-1));
```
kevinlokewy: This line is quite confusing at first glance, maybe we can reduce it into a method call? getTaskNumber() or something similar

src/main/java/Chauncey.java
```diff
@@ -0,0 +1,123 @@
...
+        else if (command.startsWith("mark")) {
+            markTask(command);
+        }
+        else if (command.startsWith("unmark")) {
+            unmarkTask(command);
+        }
+        else{
+            System.out.println("Invalid command.");
+        }
+    }
```
kevinlokewy: According to the coding standards on the CS2113 website the if-else statements should be like this: 
if (condition) {
    statements;
} else if (condition) {
    statements;
} else {
    statements;
}

Ekko-Technology: Yes I agree with kevinlokewy that there is a minor structural difference  in the coding format

src/main/java/Chauncey.java
```diff
@@ -0,0 +1,123 @@
...
+    private static int numOfTask = 0;
+
+    public static void printLine() {
+        System.out.println("____________________________________________________________");
+    }
+
+    private static void addTask() {
+        System.out.print("What type of task do you want to add? todo/deadline/event?");
+        Scanner in = new Scanner(System.in);
+        String type = in.nextLine();
```
Ekko-Technology: Consider using <string>.tolowercase() to make the scanned input case insensitive

src/main/java/Chauncey.java
```diff
@@ -0,0 +1,123 @@
...
+        System.out.println("Please enter the task details (split details by '/'): ");
+        String task = in.nextLine();
+        switch (type) {
```
Ekko-Technology: It is great that the case switches here follow proper CS2113 Java coding conventions

src/main/java/Todo.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public Todo(String description) {
+        super(description);
+    }
+
+    public void outputTaskDetails() {
+        System.out.println("[" + LABEL + "][" + super.getStatusIcon() + "] " + super.getDescription());
+    }
+
+    public String getTaskDetails() {
```
Ekko-Technology, line 22: Consider adding the @overwrite tag when using superclasses and overwriting methods from your parent class

Ekko-Technology (commented):
Looks good to me!
Ensure that Java coding standards are adopted for if-else statements.

src/main/java/Chauncey/exception/ChaunceyException.java
```diff
@@ -0,0 +1,15 @@
...
+    }
+
+    public ChaunceyException(String message) {
+        super(message);
+    }
+
+    public ChaunceyException(String message, Throwable cause) {
+        super(message, cause);
+    }
+}
```
Daviancold, line 15: Good use of overloading! Might be better to use a name to describe your exception for better readability

src/main/java/Chauncey/ui/Chauncey.java
```diff
@@ -0,0 +1,200 @@
...
+                break;
+            case "remove":
+                removeTask(command);
+                break;
+            case "mark":
+                markTask(command);
+                break;
+            case "unmark":
+                unmarkTask(command);
+                break;
```
Daviancold, line 101: Neat use of switch case, as well as putting each case logic into a function for better readability

src/main/java/Chauncey/ui/Chauncey.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    private static void unmarkTask(String command) {
+        try {
+            if (numOfTask == 0) {
+                throw new ChaunceyException("There is no task in the list. Can't do unmark command.");
+            }
+            int taskNumber = getTaskNumber(command);
+            if (!tasks[taskNumber-1].getStatus()) {
```
Daviancold: Do make sure arithmetics have a space between the symbols just as how you have written in other parts of the code
```suggestion
if (!tasks[taskNumber - 1].getStatus()) {
```

PR #17, opened 2025-08-30
[Sze Anh] iP

JeanPerrierIII:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
...
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
+    }
+}
```
msc-123456: I think it's good to use seperate files to make deadline、todo and event. Nice!

src/main/java/Jord.java
```diff
@@ -0,0 +1,114 @@
...
+            break;
+        default:
+            addTask(input);
+            break;
+        }
+    }
+
+
+    public static void main(String[] args) {
+        System.out.println("    Hello! I'm Jord\n    What can I do for you?");
```
msc-123456: I think you can add a logo like the given "Duke"? It might be more good looking this way?

src/main/java/Jord.java
```diff
@@ -0,0 +1,114 @@
...
+            addEvent(input[1]); // incomplete
+            break;
+        case "deadline" :
+            addDeadline(input[1]); // incomplete
+            break;
+        default:
+            addTask(input);
+            break;
+        }
+    }
```
msc-123456: I think you can give some advice if the input is not correct, like "please use the format ..."?

src/main/java/Jord.java
```diff
@@ -0,0 +1,114 @@
...
+
+
+    public static void main(String[] args) {
+        System.out.println("    Hello! I'm Jord\n    What can I do for you?");
+        while (true) {
+            String[] userInput = getUserInput();
+            processInput(userInput);
+        }
+    }
+
```
msc-123456: maybe you can add a method to print lines so that the intervals between each behavior can be clearer

msc-123456 (commented):
I think the general code is good, and maybe you can add some helper functions

src/main/java/Jord.java
```diff
@@ -1,23 +1,38 @@
...
         line = in.nextLine();
         while (!line.equals("bye")) {
             if (line.equals("list")) {
-                // print consecutively, list of tasks
-                for (int i = 0; i < taskCount; i++) {
-                    System.out.println("    " + (i+1) + ". " + tasks[i]);
-                }
+                printTasks(tasks);
+            }
+            else if (line.contains("mark")) {
```
Exceptional-Khoi: I think you should use line.startsWith("mark"). 
The "contains" function maybe a future error, for example, when I enter a task named "something mark"

src/main/java/Jord.java
```diff
@@ -1,23 +1,38 @@
...
+                String[] words = line.split(" ");
+                int index = Integer.parseInt(words[1]) - 1;
+                tasks[index].setMarked(!words[0].contains("un"));
+                System.out.println(words[0].contains("un") ? "    The following task hae been marked incomplete" : "    The following task has been marked complete");
```
Exceptional-Khoi: "has" not "hae". You should be careful with typing error

src/main/java/Jord.java
```diff
@@ -2,43 +2,66 @@
...
+        while (TASKS[i] != null) {
             System.out.print("    " + (i+1) + ". ");
-            tasks[i].printStatus();
+            TASKS[i].printStatus();
             i++;
         }
     }
 
+    public static void taskMarker(String input) {
+        String[] words = input.split(" ");
```
Exceptional-Khoi: split(" ") is not ok when you enter "read *two spaces* book"
you can use split("\\s+") and trim()

src/main/java/Jord.java
```diff
@@ -2,43 +2,66 @@
...
-            line = in.nextLine();
+        while (true) {
+            String userInput = getUserInput();
+            processInput(userInput);
         }
-
-        System.out.println("    Bye, see you again!");
     }
+
+
```
Exceptional-Khoi: The rest seems to be okay. I cannot find any errors or bugs. Good job on constantly improving your code based on others' feedback

src/main/java/Jord.java
```diff
@@ -0,0 +1,272 @@
...
+import java.util.Scanner;
+
+public class Jord {
+
+    private static final int TASK_LIMIT = 100;
+    public static final String WELCOME_MESSAGE = "    Hello! I'm Jord\n    What can I do for you?";
+    public static final String BYE_MESSAGE = "    Bye, see you again!";
+
+    private static Task[] TASKS = new Task[TASK_LIMIT];
+    private static int TASK_COUNT = 0;
```
Daviancold: Full case variable names should be used with non-mutable variables, perhaps consider the following changes 
```suggestion
private static int taskCount = 0;
```

src/main/java/Jord.java
```diff
@@ -0,0 +1,272 @@
...
+
+    public static boolean isTaskMarkerInputValid(String[] input) {
+        if (input.length < 2 || input[1].trim().isEmpty() ) {
+            System.out.println("    Error: missing task index");
+            return false;
+        }
+        // task index is present, and not empty
+
+        try {
+            int i = Integer.parseInt(input[1]);
```
Daviancold: The variable i here is not being used as an iterable, perhaps consider using a name or not at all?

PR #24, opened 2025-08-31
[JianHao24] iP

JianHao24:


src/main/java/Coach.java
```diff
@@ -14,25 +19,39 @@ public static void main(String[] args) {
 
         while (true) {
             String input = sc.nextLine().trim();
+            String[] inputs = input.split(" ");
 
-            if (input.equalsIgnoreCase("bye")) {
+            if (inputs[0].equalsIgnoreCase("bye")) {
                 System.out.println("____________________________________________________________");
```
lukeai-tan, line 25: Maybe you could consider converting this into a printLine() function? It could make it easier to reuse throughout your program.

src/main/java/Coach.java
```diff
@@ -1,9 +1,14 @@
 import java.util.Scanner;
 
 public class Coach {
-    static String[] tasks = new String[100];
+    static Task[] tasks = new Task[100];
     static int taskCount = 0;
 
+    private static String printTask(Task task) {
```
lukeai-tan, line 7: You could consider putting this in within the Task class and overloading the toString() function

lukeai-tan (changes requested):
Overall LGTM!

src/main/java/Coach.java
```diff
@@ -0,0 +1,64 @@
...
+        return mark + " " + task.getName();
+    }
+
+    public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
+
+        System.out.println("____________________________________________________________");
+        System.out.println(" Hello! I'm Coach");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
```
lukeai-tan: Maybe you could consider converting this into a printLine() function? It could make it easier to reuse throughout your program.

src/main/java/Coach.java
```diff
@@ -0,0 +1,64 @@
+import java.util.Scanner;
+
+public class Coach {
+    static Task[] tasks = new Task[100];
+    static int taskCount = 0;
+
+    private static String printTask(Task task) {
+        String mark = task.isDone() ? "[X]" : "[ ]";
+        return mark + " " + task.getName();
+    }
```
lukeai-tan: You could consider putting printTask within the Task class' toString function itself. You might find it easier to inherit and overwrite said toString() from Task when adding Todo, Event and Deadline

lukeai-tan (commented):
Overall LGTM.

src/main/java/Coach.java
```diff
@@ -14,25 +19,39 @@ public static void main(String[] args) {
 
         while (true) {
             String input = sc.nextLine().trim();
+            String[] inputs = input.split(" ");
 
-            if (input.equalsIgnoreCase("bye")) {
+            if (inputs[0].equalsIgnoreCase("bye")) {
                 System.out.println("____________________________________________________________");
                 System.out.println(" Bye. Hope to see you again soon!");
```
Tanjy55: Consider implementing SLAP technique so that main method is not handling so many actions.

src/main/java/Coach.java
```diff
@@ -14,25 +19,39 @@ public static void main(String[] args) {
...
                 System.out.println("____________________________________________________________");
                 break;
-            }
-            else if (input.equalsIgnoreCase("list")) {
+            } else if (inputs[0].equalsIgnoreCase("list")) {
                 System.out.println("____________________________________________________________");
+                System.out.println(" Here are the tasks in your list:");
                 for (int i = 0; i < taskCount; i++) {
```
Tanjy55: If statement/loop nesting level is beginning to get high, can these be extracted as methods to make the code more readable? Remember to follow code quality naming conventions.

src/main/java/Coach.java
```diff
@@ -1,9 +1,14 @@
 import java.util.Scanner;
 
 public class Coach {
-    static String[] tasks = new String[100];
+    static Task[] tasks = new Task[100];
```
Tanjy55: This works, but consider if a non static array size could be beneficial.

src/main/java/Coach.java
```diff
@@ -1,9 +1,14 @@
 import java.util.Scanner;
 
 public class Coach {
```
Tanjy55: Coach is an intuitive name and related well to the task tracking function currently implemented.

Tanjy55 (commented):
Good base so far!

src/main/java/Coach.java
noellethen: I agree with the user above! Should there be a constant for the maximum array size instead?

src/main/java/Coach.java
```diff
@@ -0,0 +1,210 @@
...
+
+            case "event":
+                handleEvent(input);
+                break;
+
+            default:
+                printLine();
+                System.out.println("Invalid command!");
+                printLine();
+                break;
```
noellethen: I like how you didn't indent the case statements.

src/main/java/Coach.java
```diff
@@ -0,0 +1,210 @@
...
+                return;
+            }
+
+            tasks[Task.getTaskCount()] = new Event(taskName, fromTime, toTime);
+            Task.incrementTaskCount();
+            printAddTask();
+
+        } catch (Exception e) {
+            printInvalidFormat();
+            return;
```
noellethen: I like that the error handling in each function is very comprehensive!

src/main/java/Event.java
```diff
@@ -0,0 +1,37 @@
...
+        super(name, false);
+        this.from = from;
+        this.to = to;
+    }
+
+    public Event(String name, boolean isDone, String from, String to) {
+        super(name, isDone);
+        this.from = from;
+        this.to = to;
+    }
```
noellethen, line 59: Is there a reason why there's another constructor?

noellethen (commented):
Overall, I like your code! It's very clean and readable as it's broken down into many functions. Good job!

src/main/java/Coach.java
shennontay: agree with the above, to add on, maybe the reference tasks can also be final so that it is not mutable

src/main/java/Task.java
```diff
@@ -0,0 +1,57 @@
...
+    public static int getTaskCount() {
+        return taskCount;
+    }
+
+    public static void setTaskCount(int taskCount) {
+        Task.taskCount = taskCount;
+    }
+
+    public static void incrementTaskCount() {
+        Task.taskCount = taskCount++;
```
shennontay: i think the assignment happens before the increment so the increment might not happen?

src/main/java/Coach.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    private static void handleToDo(String input) {
+        printLine();
+        String content = input.substring(4);
+        if (content.isEmpty())
+        {
+            printEmpty();
+            return;
+        }
```
shennontay: you can put } and else on the same line 

```suggestion
        } else {
```

src/main/java/Coach.java
shennontay: this is so nice and readable :))

shennontay (commented):
readable and good error handling yay nice work

src/main/java/chatbot/Coach.java
```diff
@@ -0,0 +1,203 @@
...
+
+import chatbot.tasks.Deadline;
+import chatbot.tasks.Event;
+import chatbot.tasks.Task;
+import chatbot.tasks.ToDo;
+
+import java.util.Scanner;
+
+public class Coach {
+    public static final int MAX_TASKS = 100;
```
namsengi11: Good job defining hardcoded outputs as constants

src/main/java/chatbot/Coach.java
```diff
@@ -0,0 +1,203 @@
...
+
+        while (true) {
+            String input = sc.nextLine().trim();
+            String[] inputs = input.split(" ");
+            String command = inputs[0].toLowerCase();
+
+            try {
+                switch (command) {
+                case "bye":
+                    handleBye();
```
namsengi11: Good job abstracting out the logic of each case by using other methods. Makes the flow more readable.

src/main/java/chatbot/Coach.java
```diff
@@ -0,0 +1,203 @@
...
+            } catch (CoachException e) {
+                System.out.println(e.getMessage());
+                printLine();
+            }
+        }
+    }
+
+    private static void handleToDo(String input) throws CoachException {
+        printLine();
+        String content = input.substring(4);
```
namsengi11: The size of each offset for the substring slicing is not immediately obvious, could consider refactoring them as constants.

src/main/java/chatbot/tasks/Task.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public int getId() {
+        return id;
+    }
+
+    public Task(String name, boolean isDone) {
+        this.name = name;
+        this.isDone = isDone;
+        taskCount++;
```
namsengi11: Is the taskcount being incremented twice, as it is being incremented in constructor and incrementTaskCount() method?

PR #31, opened 2025-08-31
[ChangIkJoong] iP

ChangIkJoong:
First Pull Request as IP assignment's requirements. Hopefully this works!!


src/main/java/Controller.java
```diff
@@ -0,0 +1,97 @@
...
+            unmarkTaskDone(inputPrompt);
+        } else if (inputPrompt.equalsIgnoreCase("list")) {
+            view.viewTaskList(model.getTasks());
+        } else if (inputPrompt.toLowerCase().startsWith("deadline ")) {
+            addDeadlineTask(inputPrompt);
+        } else if (inputPrompt.toLowerCase().startsWith("event ")) {
+            addEventTask(inputPrompt);
+        } else {
+            addTodoTask(inputPrompt);
+        }
```
JeanPerrierIII: maybe try splitting the input prompt into words, then just look at the first word and make switch case

src/main/java/Model.java
```diff
@@ -0,0 +1,39 @@
...
+    public Task getTask(int index) {
+        return tasks.get(index);
+    }
+
+    public ArrayList<Task> getTasks() {
+        return tasks;
+    }
+
+    public boolean markDone(int index) {
+        if (index >= 0 && index < tasks.size()) {
```
JeanPerrierIII: Might be able to extract the range check out for even cleaner code

src/main/java/View.java
```diff
@@ -0,0 +1,78 @@
...
+            resultString = "[T]" + isCompleteStatus(task) + task.getTaskDescription();
+        } else if (task instanceof Deadline) {
+            Deadline deadline = (Deadline) task;
+            resultString = "[D]" + isCompleteStatus(task) + deadline.getTaskDescription() + " (by: " + deadline.getEndDate() + ")";
+        } else if (task instanceof Event) {
+            Event event = (Event) task;
+            resultString = "[E]" + isCompleteStatus(task) + event.getTaskDescription() + " (from: " + event.getStartDate() + " to: " + event.getEndDate() + ")";
+        } else {
+            resultString = isCompleteStatus(task) + task.getTaskDescription();
+        }
```
JeanPerrierIII: Might be able to extract these into their own functions within the class for cleaner code, maybe convert the else-ifs to a switch case as well

src/main/java/Model.java
JeanPerrierIII, line 1: overall really like how clean the class is

JeanPerrierIII (commented):
Overall great code, nice and easy to read and understand 👍
Some places could possible be neatened up abit but still good.

src/main/java/Controller.java
```diff
@@ -0,0 +1,97 @@
...
+        if (parts.length == 3) {
+            Task newInput = new Event(parts[0], parts[1], parts[2]);
+            model.addTask(newInput);
+            view.viewTaskAdded(newInput, model.getTaskCount());
+        } else {
+            view.viewError();
+        }
+    }
+
+    private void handleInput(String inputPrompt) {
```
JianHao24: Perhaps a switch case would be better than using if and else!

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
+public abstract class Task {
+    private String taskDescription;
+    private boolean isDone;
+    private int taskId;
+    private static int counter = 1;
```
JianHao24: can have a better name than counter like taskCounter which is more specific and non generic

src/main/java/TaskView.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public void viewTaskAdded(Task task, int total) {
+        System.out.println("____________________________________________________________");
+        System.out.println(" Got it. I've added this task:");
+        System.out.println("   " + viewTask(task));
+        System.out.println(" Now you have " + total + " tasks in the list.");
+    }
+
+    public void viewExit() {
+        System.out.println("Bye.");
```
Exceptional-Khoi: You can use System.exit() to completely stop the running program

src/main/java/Todo.java
```diff
@@ -0,0 +1,5 @@
+public class Todo extends Task {
+    public Todo(String taskDescription) {
+        super(taskDescription);
+    }
+}
```
Exceptional-Khoi, line 7: The program is well-orgranized. You have many classes, but each class has needed function and does its job great. It creates a good feeling for readers and reviewers, because we don't have to read everything in "Bruce"

src/main/java/Bruce.java
```diff
@@ -4,83 +4,102 @@
...
     }
 
-    public static String inputFromUser() {
+    private static String inputFromUser() {
         return inputFromKeyboard.nextLine();
     }
 
-    public static void markDone(String prompt) {
+    private static void markDone(String prompt) {
         String[] parts = prompt.split(" ");
```
Exceptional-Khoi: split(" ") doesn't work for the case like: "read *two spaces* book". you can use trim()

src/main/java/Bruce.java
```diff
@@ -4,83 +4,102 @@
...
+            newDeadlineTask(inputPrompt);
+        } else if (inputPrompt.toLowerCase().startsWith("event ")) {
+            newEventTask(inputPrompt);
         } else {
-            Task newInput = new Task(inputPrompt);
-            tasks.add(newInput);
-            System.out.println("added: " + inputPrompt);
+            newTodoTask(inputPrompt);
         }
     }
```
Exceptional-Khoi: Bruce class is a little too long. If you add more functions in the future, it'll be even longer. You can move some functions to TaskView class

src/main/java/View.java
```diff
@@ -0,0 +1,78 @@
+public class View {
+    public static final String INLINE_TEXT_LINES = "____________________________________________________________";
```
JianHao24: Constants are set in SCREAMING_SNAKE_CASE which follows convention. Good Job!

src/main/java/View.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public void viewTaskAdded(Task task, int total) {
+        System.out.println("____________________________________________________________");
+        System.out.println(" Got it. I've added this task:");
+        System.out.println("   " + viewTask(task));
+        System.out.println(" Now you have " + total + " tasks in the list.");
+    }
+
+    public void viewExit() {
```
JianHao24: Instead of hardcoding "Bye." or "Error. Please try again.", instead define them as constants at the top like (private static final String DEFAULT_ERROR_MESSAGE = "Error. Please try again."; ). This would increase the readability of your code and reduce redudancy when printing the same error message.

src/main/java/Model.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public Task getTask(int index) {
+        return tasks.get(index);
+    }
+
+    public ArrayList<Task> getTasks() {
+        return tasks;
+    }
+
+    public boolean markDone(int index) {
```
JianHao24: boolean can be isDone to better follow Java conventionand named to sound like a boolean instead of markDone

ChangIkJoong: The line you are reffering to is a method not a boolean variable. According to conventions this should not be a problem. Please correct me if I'm wrong, thank you :) !

src/main/java/View.java
ChangIkJoong: As of week 4, we had not yet discussed exceptions and error throwings, this is why I hard coded it in order to not jump ahead, but thank you for great feedback and will implement this in further iteration! :D

src/main/java/Task.java
ChangIkJoong: Nice catch, thank you! :D

src/main/java/UI/View.java
```diff
@@ -0,0 +1,99 @@
...
+        System.out.println("  todo <description>                    : Adds a Todo task.");
+        System.out.println("  deadline <desc> /by <date>            : Adds a Deadline task with a due date.");
+        System.out.println("  event <desc> /from <start> /to <end>  : Adds an Event task with start and end times.");
+        System.out.println("  list                                  : Shows all tasks.");
+        System.out.println("  mark <task number>                    : Marks a task as done.");
+        System.out.println("  unmark <task number>                  : Marks a task as not done.");
+        System.out.println("  bye.                                  : Exits the chatbot.");
+        System.out.println("  help                                  : Shows this instruction list.");
+    }
+}
```
JoanneHing, line 188: Great job setting up the view architecture for iP. It's a good application of what covered in the lecture.

src/main/java/Bruce/Controller.java
```diff
@@ -0,0 +1,140 @@
...
+import Model.TaskVariants.Task;
+import Model.TaskVariants.Todo;
+import UI.View;
+
+import java.util.Scanner;
+
+public class Controller {
+    private final Model model;
+    private final View view;
+    private boolean isRunning = true;
```
JoanneHing, line 28: Good job in naming the boolean following the coding standards.

src/main/java/Bruce/Controller.java
```diff
@@ -0,0 +1,140 @@
...
+                view.viewError("Missing task number. Example: \"mark 3\"");
+            } catch (DukeException e) {
+                view.viewError(e.getMessage());
+            } catch (Exception e) {
+                view.viewError("Unexpected error: " + e.getMessage());
+            } finally {
+                View.printLine();
+            }
+        }
+    }
```
JoanneHing, line 68: Good to see that you include the finally block in a try-catch block, its a good practice!

src/main/java/Bruce/Controller.java
```diff
@@ -0,0 +1,140 @@
...
+        switch (command) {
+        case "bye." -> exitProgram();
+        case "mark" -> markTaskDone(inputPrompt);
+        case "unmark" -> unmarkTaskDone(inputPrompt);
+        case "list" -> view.viewTaskList(model.getTasks());
+        case "deadline" -> addDeadlineTask(inputPrompt);
+        case "event" -> addEventTask(inputPrompt);
+        case "todo" -> addTodoTask(inputPrompt);
+        case "help" -> View.showInstructions();
+        default -> throw new UnknownInputException(command);
```
JoanneHing, line 257: It is allowed to use lambda blocks, but remember to indent the `case` blocks, as the coding standard would suggest lambda `switch-case` to have indented `case`. May review the coding standard page for related reference for "Lambda-style switch statements/expressions can have indented case blocks"

src/main/java/Model/Model.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public int getTaskCount() {
+        return tasks.size();
+    }
+
+    private boolean rangeCheck(int index) {
+        return (index >= 0 && index < tasks.size());
+    }
+}
```
JoanneHing, line 145: Good job on implementing the Model architecture too.

src/main/java/Model/TaskVariants/Deadline.java
```diff
@@ -0,0 +1,20 @@
+package Model.TaskVariants;
+
+public class Deadline extends Task {
+    protected String endDate;
```
JoanneHing, line 4: May I clarify why you decided to use `protected` instead of `private` variables? Would you have subclass that would need to use this variable?

src/main/java/Model/TaskVariants/Event.java
```diff
@@ -0,0 +1,30 @@
+package Model.TaskVariants;
+
+public class Event extends Task {
+    protected String startDate;
+    protected String endDate;
```
JoanneHing, line 5: Same as the comment in `Deadline.java`

JoanneHing (commented):
Overall well done! Good job to implement MVC architecture that is taught in the lecture too.

src/main/java/UI/View.java
JoanneHing: Missed out on this, instead using lots of `get` method to get the detail of the tasks, maybe you can consider overwritting the `toString()` method of the task classes better?

PR #32, opened 2025-09-01
[shauntsr] iP

shauntsr:


src/main/java/Odin.java
```diff
@@ -0,0 +1,44 @@
+import java.util.Scanner;
+
+public class Odin {
+    public static void printHorizontalLine(){
```
sanjay-shiva-kumar: I like how you made printing horizontal line a separate function 👍

src/main/java/TaskList.java
```diff
@@ -0,0 +1,28 @@
...
+    public void unmarkTask(int taskIndex) {
+        taskList[taskIndex].setIsDone(false);
+        System.out.println("Please tell me it was a misinput.");
+        System.out.println("   " + taskList[taskIndex].getDisplayString());
+    }
+
+    public void printTasks() {
+        for (int i = 0; i < taskCount; i++) {
+            Task currTask = taskList[i];
+            System.out.println(i+1 + ". " + currTask.getDisplayString());
```
Wrooi: Any reason why your "i+1" not properly spaced out according to Java coding standard?

src/main/java/Odin.java
```diff
@@ -0,0 +1,44 @@
...
+            String input = scanner.nextLine();
+
+            printHorizontalLine();
+
+            if (input.equals("bye")) {
+                System.out.println("Bye. Hope to see you again soon!");
+                printHorizontalLine();
+                return;
+            }
+            if (input.equals("list")) {
```
Wrooi: Any reason why your "if (input.equals("list")) {" an else if instead of an if?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,28 @@
+public class TaskList {
```
Wrooi: I like how you had multiple classes for different purposes with self-explanatory names 👍

src/main/java/TaskList.java
```diff
@@ -0,0 +1,28 @@
...
+        taskList[taskIndex].setIsDone(true);
+        System.out.println("Nice! Another task down!");
+        System.out.println("   " + taskList[taskIndex].getDisplayString());
+    }
+
+    public void unmarkTask(int taskIndex) {
+        taskList[taskIndex].setIsDone(false);
+        System.out.println("Please tell me it was a misinput.");
+        System.out.println("   " + taskList[taskIndex].getDisplayString());
+    }
```
Wrooi: Perhaps you could combine this as one toggle function?

src/main/java/Task.java
```diff
@@ -0,0 +1,24 @@
...
+        return (isDone ? "[X]" : "[ ]");
+    }
+
+    public void setIsDone(boolean isDone) {
+        this.isDone = isDone;
+    }
+
+    public String getDisplayString() {
+        return getStatusIcon() + " " + description;
+    }
```
Wrooi, line 67: The Methods are correctly named in camelCase.👍

src/main/java/Odin.java
```diff
@@ -0,0 +1,44 @@
+import java.util.Scanner;
+
+public class Odin {
+    public static void printHorizontalLine(){
+        System.out.println("__________________________________________");
+    }
+    public static void main(String[] args) {
+        TaskList t = new TaskList();
```
Wrooi: Could there be a more intuitive name instead of t?

src/main/java/Task.java
```diff
@@ -0,0 +1,24 @@
+public class Task {
```
Wrooi: The Classes are correctly named in PascalCase.👍

Wrooi (commented):
Good job

src/main/java/TaskList.java
shauntsr: Thank you for noting this, I have amended accordingly.

src/main/java/Odin.java
namsengi11: Requires space between () and {

src/main/java/TaskList.java
```diff
@@ -0,0 +1,29 @@
+public class TaskList {
+    private Task[] taskList = new Task[100];
```
namsengi11: Consider factoring out 100 as a constant, as it is a magic number.

PR #34, opened 2025-09-01
[gohjieling834] iP

gohjieling834:


src/main/java/Berry.java
```diff
@@ -0,0 +1,57 @@
+import java.util.Scanner;
+
+public class Berry {
+    public static void main(String[] args) {
+        String chatbotName = "Berry";
+        String userCommand;
+        Task[] tasks = new Task[100];   //array of Task objects
```
saheer17: Great that you used plural form for tasks, which is a collection of objects! Instead of using the number "100", which is a magic number, you can add a int variable named "taskCapacity" and initialise it as 100, then use that variable when initialising your array.

src/main/java/Berry.java
```diff
@@ -0,0 +1,57 @@
+import java.util.Scanner;
+
+public class Berry {
+    public static void main(String[] args) {
+        String chatbotName = "Berry";
```
saheer17: Since "Berry" is a constant, you can mark the variable as "final" and for constants, change the name of the variable such that it is in UPPER_SNAKE_CASE form.

src/main/java/Berry.java
```diff
@@ -0,0 +1,57 @@
...
+            if (userCommand.equals("list")) {
+                printList(tasks, numberOfTasks);
+            } else if (userCommand.contains("mark")) {
+                toggleTaskStatus(userCommand, tasks);
+            } else {
+                addTask(userCommand, tasks, numberOfTasks);
+                numberOfTasks++;
+            }
+            userCommand = in.nextLine();
+        }
```
saheer17: For better readability, consider replacing the chained if–else statements with a switch on userCommand. This makes the command handling easier to read and easier to extend if new commands are added later

src/main/java/Task.java
```diff
@@ -0,0 +1,37 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
saheer17, line 5: The variables are declared as "protected", which can expose them and break encapsulation. You may want to declare them as private instead

saheer17 (commented):
Overall, Java coding standards were mostly followed, especially in the case of naming variables. You may want to consider using switch cases more to improve readability!

src/main/java/Task.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
+    }
+
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    public void unmarkAsDone() {
```
yujie-o:
```suggestion
    public void markAsUndone() {
```
Prefer intention-revealing methods for state changes. For naming, use verbs for actions.

src/main/java/Berry.java
```diff
@@ -0,0 +1,57 @@
...
+        System.out.println("\n‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧\nadded: " + command + "\n‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧\n");
+    }
+
+    public static void printList(Task[] tasks, int numberOftasks) {
+        System.out.println("\n‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧");
+        for (int i = 0; i < numberOftasks; i++) {
+            System.out.println(i + 1 + ".[" + tasks[i].getStatusIcon() + "] " + tasks[i].getDescription());
+        }
+        System.out.println("‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧\n");
+    }
```
yujie-o: Good use of spaces around operators & in string concatenations.

src/main/java/Berry.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public static void addTask(String command, Task[] tasks, int numberOftasks) {
+        tasks[numberOftasks] = new Task(command);
+        System.out.println("\n‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧\nadded: " + command + "\n‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧\n");
+    }
+
+    public static void printList(Task[] tasks, int numberOftasks) {
+        System.out.println("\n‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧");
+        for (int i = 0; i < numberOftasks; i++) {
```
yujie-o:
```suggestion
        for (int i = 0; i < numberOfTasks; i++) {
```
The variable numberOftasks doesn’t follow Java naming conventions. Variable names should be in camelCase (so it should be numberOfTasks), and in this case, taskCount or simply size would be clearer and more consistent with Java’s collection classes.

src/main/java/Task.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public boolean isDone() {
+        return isDone;
+    }
+
+    public void setDone(boolean done) {
+        isDone = done;
+    }
+
+    public String getStatusIcon() {
```
yujie-o, line 24: The isDone boolean and getStatusIcon() are simple but effective abstractions for tracking task state.

yujie-o (commented):
Overall, this is a solid start with good structure, a few adjustments around naming, validation, and documentation will make the code much more robust and aligned with Java coding standards.

src/main/java/Berry.java
```diff
@@ -0,0 +1,147 @@
+import java.util.Scanner;
+
+public class Berry {
+    public static final String DIVIDER = "‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧‧⋆ ✧˚₊‧⋆. ✧˚₊‧⋆‧";
```
W1ndB10w: This divider pattern is beautifully designed.

src/main/java/Berry.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public static void printList(Task[] tasks, int taskCount) {
+        System.out.println("\n" + DIVIDER);
+        for (int i = 0; i < taskCount; i++) {
+            System.out.println(i + 1 + "." + tasks[i]);
+        }
+        System.out.println(DIVIDER + "\n");
+    }
+
+    public static void toggleTaskStatus(String userInput, Task[] tasks) {
```
W1ndB10w: You consistently used CamelCase in function names and variable names.
They are very clear for user to read and understand their usages. Good job!

src/main/java/Berry.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public static String[] extractDetails(String userInput) {
+        int dividerPosition = userInput.indexOf(" ");
+        String task = userInput.substring(dividerPosition + 1);
+        return task.split("/");
+    }
+
+    public static int executeCommand(String userInput, Task[] tasks, int taskCount) {
+        String userCommand = extractCommand(userInput);
+        switch (userCommand) {
```
W1ndB10w: I like the way that you used "switch-case" when writing your code.
They can convey your logic more clearly.
However, you can also consider adding a "**default**" section to make the structure more complete.
You also used the correct coding format. Good job!

W1ndB10w (commented):
Generally clear coding format and good usage of "switch-case".

src/main/java/berry/Berry.java
```diff
@@ -0,0 +1,179 @@
...
+    public static final Scanner IN = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        String userInput;
+        Task[] tasks = new Task[TASK_CAPACITY];   //array of Task objects
+        int taskCount = 0;
+
+        printHelloMessage();
+        while (true) {
+            userInput = getUserInput(IN);   //get command from user
```
roshnidaksha: Avoid giving comments to code that is self-explanatory.

src/main/java/berry/task/Task.java
```diff
@@ -0,0 +1,44 @@
+package berry.task;
+
+public class Task {
```
roshnidaksha: Since Task itself is never directly instantiated and objects are always Todo, Deadline, or Event, it would be better to declare Task as abstract.

src/main/java/berry/Berry.java
```diff
@@ -0,0 +1,179 @@
...
+                System.out.println("\n" + DIVIDER + "\n" + "Nice! I've marked this task as done:\n  "
+                        + tasks[taskNumber] + "\n" + DIVIDER + "\n");
+            }
+        } catch (NumberFormatException e) {
+            printErrorMessage("Sorry, I don't know which task to delete. Please enter the task number, thank you! :)");
+        } catch (NullPointerException e) {
+            printErrorMessage("This task number does not exist! :|");
+        }
+    }
+}
```
roshnidaksha, line 53: You can consider extracting related data/methods into separate classes to make this class shorter.

roshnidaksha (commented):
There were a few obvious comments mentioned in the code.
Otherwise, well written and neat code!

PR #44, opened 2025-09-01
[hyeungac] ip

hyeungac:
okiii, so I've done increment 0-3, now what?


src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
...
+    public void setName(String name) {
+        this.name = name;
+    }
+    public boolean isDone() {
+        return done;
+    }
+    public void setDone(boolean done) {
+        this.done = done;
+    }
+    public void printinfo(){
```
shira421: Should this be named printInfo?

shira421: Should there be a space between () and { ?

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
...
+        System.out.print("[" );
+        if (this.done){
+            System.out.print("X");
+        }
+        else{
```
shira421, line 34: Perhaps the else statement should be on the same line as the closing brace for the if statement.

src/main/java/Todo.java
```diff
@@ -0,0 +1,12 @@
+public class Todo extends Task{
+    public Todo(String name){
+        super(name);
+    }
+
+    @Override
```
shira421, line 12: I like how you used the @/Override tag to indicate overriding of the function.

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
+public class Task {
+    private String name;
+    private boolean done;
```
shira421, line 7: Perhaps this variable should be called isDone?

shira421 (commented):
LGTM. Just a few nits to fix.

src/main/java/Duke.java
limeiy1, line 1: maybe you can define a constant for the horizontal line instead of writing out the line every time you need to print it?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,17 @@
...
+    public Deadline(String name, String by){
+        super(name);
+        this.by = by;
+    }
+    public String getBy() {
+        return by;
+    }
+
+    @Override
+    public void printinfo() {
```
limeiy1: I like how you used a verb for the method name here which makes the method clear, but perhaps you can write it in camelCase? I think the same applies to some of your other method names

src/main/java/Duke.java
```diff
@@ -1,10 +1,92 @@
...
+                    String TaskName = line.substring(firstspace + 1);
+                    list[task_count++] = new Todo(TaskName);
+                }
+                else if(line.contains("deadline")){
+                    String[] parts = line.split("/");
+                    int firstspace = parts[0].indexOf(" ");
+                    String TaskName = parts[0].substring(firstspace + 1);
+                    list[task_count++] = new Deadline(TaskName, parts[1]);
+                }
+                else if (line.contains("event")){
```
limeiy1: maybe you can standardise the spacing between "if" and the bracket "("

src/main/java/Duke.java
```diff
@@ -1,10 +1,92 @@
...
+        while (true) {
+            String line;
+            Scanner in = new Scanner(System.in);
+            line = in.nextLine();
+            if(line.equalsIgnoreCase("bye")){
+                System.out.println("____________________________________________________________\n" +
+                        " Bye. Hope to see you again soon!\n" +
+                        "____________________________________________________________\n");
+                break;
+            }
```
limeiy1: I like how you included a blank line between these two sections to structure your code more logically! Perhaps you could do the same for other sections of your code?

limeiy1 (commented):
Overall, I think your code is easy to read and understand with the use of clearly defined methods and clear if-else statements.

PR #49, opened 2025-09-01
[argadde] iP

argadde:


src/main/java/Bevo.java
```diff
@@ -0,0 +1,206 @@
...
+
+public class Bevo {
+    /** List of commands available to the user */
+    private static final String BYE_COMMAND = "bye";
+    private static final String LIST_COMMAND = "list";
+    private static final String MARK_COMMAND = "mark";
+    private static final String UNMARK_COMMAND = "unmark";
+    private static final String TODO_COMMAND = "todo";
+    private static final String DEADLINE_COMMAND = "deadline";
+    private static final String EVENT_COMMAND = "event";
```
mendax1234: Good job for declaring the Constants as `private static final`.

src/main/java/Task.java
```diff
@@ -0,0 +1,51 @@
...
+     * isDone is set to false by default.
+     * 
+     * @param description Description of task to be completed
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    /**
```
mendax1234, line 28: Good job for doing the JavaDoc style explanation for the method!

src/main/java/Todo.java
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * Displays a String that contains the status and
+     * description of the Todo.
+     */
+    @Override
+    public String toString() {
+        return "[T]" + super.toString();
+    }
+}
```
mendax1234, line 30: Good job for using the inheritence! It makes the structure clearer

src/main/java/Bevo.java
```diff
@@ -0,0 +1,206 @@
...
+            } else if (input.toLowerCase().startsWith(EVENT_COMMAND)) {
+                taskCount = executeEventCommand(tasks, taskCount, input);
+            } else {
+                tasks[taskCount] = new Todo(input);
+                printAddCommand(tasks[taskCount], ++taskCount);
+            }
+        }
+
+        scanner.close();
+    }
```
mendax1234: Any reason why you did this in the main method instead of creating another class maybe called `Parser.java` for parsing the input?

src/main/java/Bevo.java
```diff
@@ -0,0 +1,206 @@
...
+            } else if (input.toLowerCase().startsWith(TODO_COMMAND)) {
+                taskCount = executeToDoCommand(tasks, taskCount, input);
+            } else if (input.toLowerCase().startsWith(DEADLINE_COMMAND)) {
+                taskCount = executeDeadlineCommand(tasks, taskCount, input);
+            } else if (input.toLowerCase().startsWith(EVENT_COMMAND)) {
+                taskCount = executeEventCommand(tasks, taskCount, input);
+            } else {
+                tasks[taskCount] = new Todo(input);
+                printAddCommand(tasks[taskCount], ++taskCount);
+            }
```
HarshitSrivastavaHS: Since you're converting input to lowerCase in all conditions, can declare a variable before the block and do it just once (inputLowerCase or something)

src/main/java/Bevo.java
```diff
@@ -0,0 +1,206 @@
...
+     * @param input the user's input
+     */
+    private static void executeUnmarkCommand(Task[] tasks, String input) {
+        String taskNumber = input.split(" ")[1];
+        int index = Integer.parseInt(taskNumber) - 1;
+        tasks[index].markAsNotDone();
+        System.out.println(HORIZONTAL_LINE);
+        System.out.println("\t  OK, I've marked this task as not done yet:");
+        System.out.println("\t\t  " + tasks[index]);
+        System.out.println(HORIZONTAL_LINE + "\n");
```
HarshitSrivastavaHS: You're printing horizontal line before and after each command. Can just make a new function to do this for you (maybe a function where you just pass the value to be printed and it prints the value with 2 horizontal lines, one at the start and one at the end)

src/main/java/Bevo.java
HarshitSrivastavaHS: Can also consider using switch case instead of if block.

src/main/java/Bevo.java
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * Prints out a welcome message with the chatbot's
+     * name and prompts the user for input.
+     */
+    private static void printWelcomeMessage() {
+        String logo = "|||||   -----   \\     /   -----\n"
+                + "|    |  |        \\   /   |     |\n"
+                + "|||||   ----      \\ /    |     |\n"
+                + "|    |  |          |     |     |\n"
+                + "|||||   -----      |      ----- \n";
```
HarshitSrivastavaHS: Since the logo wouldn't be changing, you can just declare it as final at the top of the class with other constants.

HarshitSrivastavaHS:
Very well written and organised code :))

src/main/java/Event.java
```diff
@@ -0,0 +1,32 @@
+public class Event extends Task {
+    
+    /** Start time of the timeline */
+    protected String from;
```
roshnidaksha, line 4: You don't need to give comments to code that is self-explanatory.

src/main/java/Task.java
```diff
@@ -0,0 +1,51 @@
+public class Task {
```
roshnidaksha: Since Task itself is never directly instantiated and objects are always Todo, Deadline, or Event, it would be better to declare Task as abstract. This makes the design clearer and prevents accidental creation of plain Task objects.

src/main/java/Bevo.java
```diff
@@ -0,0 +1,206 @@
...
+    private static int executeEventCommand(Task[] tasks, int taskCount, String input) {
+        String[] parts = input.substring(EVENT_COMMAND.length() + 1).split("/from|/to");
+        String description = parts[0].trim();
+        String from = parts.length > 1 ? parts[1].trim() : UNSPECIFIED_MESSAGE;
+        String to = parts.length > 2 ? parts[2].trim() : UNSPECIFIED_MESSAGE;
+        tasks[taskCount] = new Event(description, from, to);
+
+        printAddCommand(tasks[taskCount], ++taskCount);
+        return taskCount;
+    }
```
roshnidaksha: The method name suggests that it should do the execution of event command. To be consistent with other execute commands, can you try finding other ways to handle the task count instead of returning it as part of this function?

src/main/java/Bevo.java
```diff
@@ -0,0 +1,206 @@
...
+                + "|    |  |          |     |     |\n"
+                + "|||||   -----      |      ----- \n";
+        System.out.println("Hello from\n" + logo);
+
+        System.out.println(HORIZONTAL_LINE);
+        System.out.println("\t  Hello! I'm Bevo.");
+        System.out.println("\t  What can I do for you?");
+        System.out.println(HORIZONTAL_LINE + "\n");
+    }
+}
```
roshnidaksha, line 40: This class is too long. Try extracting related data/methods into separate classes.

roshnidaksha (commented):
Overall, well written code. 
I have pointed some minor things in comments. Do take a look at it.

PR #50, opened 2025-09-01
[ZT712002] iP

ZT712002:


Sheeeesh-code (commented):
Good Job, you followed the coding standard for the switch

src/main/java/Passer.java
```diff
@@ -0,0 +1,111 @@
...
+    public void setPasserActive(boolean passerActive) {
+        isPasserActive = passerActive;
+    }
+
+    public void processCommand(String userInput, TaskList tasks) {
+        String[] commandParts = userInput.split(" ", 2);
+        String command = commandParts[0].toLowerCase();
+        String arguments = commandParts.length > 1 ? commandParts[1] : "";
+        try {
+            switch (command) {
```
Sheeeesh-code: Good Job, The switch follows the java coding standard

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task {
```
Sheeeesh-code, line 18: All Class should be in a packages

src/main/java/Joebot666.java
```diff
@@ -0,0 +1,17 @@
+public class Joebot666 {
```
Sheeeesh-code, line 28: Good job, you followed OOP, by making many Class for different utility, using inheritance

src/main/java/Task.java
```diff
@@ -0,0 +1,34 @@
+public class Task {
+    private String description;
+    private boolean isDone;
+
+
+    public Task(String description) {
```
Sheeeesh-code, line 15: Good job, your loop are  wrapped by curlys bracket even if there is only one line in the body

src/main/java/Passer.java
```diff
@@ -0,0 +1,111 @@
...
+                break;
+            case "event":
+                handleEvent(tasks, arguments);
+                break;
+            case "bye":
+                setPasserActive(false);
+                break;
+            }
+
+        } catch (Exception e) {
```
Sheeeesh-code, line 87: Good job, your try catch statement have the good form

src/main/java/Passer.java
```diff
@@ -0,0 +1,111 @@
+import java.util.Scanner;
+
+public class Passer {
+    private final Scanner IN;
```
Sheeeesh-code: Good job here IN is a constant hand you write it in uppercase

Sheeeesh-code (commented):
Good job, but don't forget to put all your Class in package

src/main/java/TaskList.java
```diff
@@ -0,0 +1,53 @@
+import java.util.ArrayList;
+
+public class TaskList {
```
gordonajajar: I liked that you even extracted the Tasks array into its own class, not to mention the use of ArrayList for dynamic-length arrays.

src/main/java/Passer.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public void setUserInput() {
+        this.userInput = IN.nextLine();
+    }
+
+    public boolean isPasserActive() {
+        return isPasserActive;
+    }
+
+    public void setPasserActive(boolean passerActive) {
```
gordonajajar: I'm not sure setPasserActive is a descriptive name here, since it suggests that it will set Passer to active, but it in fact takes in a boolean and sets Passer to that value. Perhaps it can be factored into 2 different functions, setPasserActive and setPasserInactive? Or just something like setPasserState

src/main/java/Passer.java
```diff
@@ -0,0 +1,111 @@
...
+        isPasserActive = passerActive;
+    }
+
+    public void processCommand(String userInput, TaskList tasks) {
+        String[] commandParts = userInput.split(" ", 2);
+        String command = commandParts[0].toLowerCase();
```
gordonajajar: Nice job pre-processing the input, even considering case

src/main/java/Passer.java
```diff
@@ -0,0 +1,111 @@
...
+            System.out.println("The description of a todo cannot be empty.");
+            return;
+        }
+        tasks.addTask(new Todo(arguments));
+    }
+
+    private void handleEvent(TaskList tasks, String arguments) {
+        String[] parts = arguments.split(" /from");
+        if (parts.length < 2) {
+            System.out.println("Please provide both description and start time for the event in this format " +
```
gordonajajar, line 208: Indenting here is off; may be fixed by setting "Autoformat on save" in IntelliJ Idea? (If that is what you are using)

gordonajajar (commented):
Nice job! You have done a lot of abstraction as seen from your short Joebot666 (main) file.

src/main/java/Deadline.java
ZT712002: Thanks for pointing this out! didn't realise that

src/main/java/Passer.java
ZT712002: Thanks for catching that. Didn't realise :)

src/main/java/Passer.java
ZT712002: Thanks, I've realised the only utility of this function would be to exit the program elegantly. I will refactor this in future commits. Thanks for pointing this out!

src/main/java/Joebot666.java
```diff
@@ -0,0 +1,20 @@
...
+    public static void main(String[] args) {
+        System.out.println("Hello I'm JoeBot666");
+        System.out.println("What can I do for you today?");
+
+        TaskList tasks = new TaskList();
+        Passer passer = new Passer();
+        while (passer.isPasserActive()) {
+            System.out.println("*************************************");
+            passer.setUserInput();
+            passer.processCommand(passer.getUserInput(), tasks);
```
Daviancold: Just a small nitpick, but perhaps you could consider storing `passer.getUserInput()` into a variable before passing it into `processCommand()`

src/main/java/exception/IllegalTaskNumberException.java
```diff
@@ -0,0 +1,4 @@
+package exception;
+
+public class IllegalTaskNumberException extends Exception {
```
Daviancold, line 6: Good and clear naming used for custom exception

PR #52, opened 2025-09-01
[aydrienlaw] iP

aydrienlaw:


src/main/java/Tilo.java
```diff
@@ -0,0 +1,112 @@
+import java.util.Scanner;
+
+public class Tilo {
+    public static final int taskListSize = 100;
+    public static final String INDENT = "\t";
```
ycdaniel326: I like the idea of defining the Indent and border line here! 
And perhaps you can consider changing taskListSize to upper case if it is a constant.

src/main/java/Tilo.java
```diff
@@ -0,0 +1,112 @@
...
+        String userInput;
+        Scanner in = new Scanner(System.in);
+        Task[] tasks = new Task[taskListSize];
+        int taskCount = 0;
+
+        do {
+            userInput = readUserInput(in);
+            printBorder();
+            String[] words = userInput.split(" ");
+            switch (words[0]) {
```
ycdaniel326: I like how you use switch statements to separate different cases. 
For the format, it seems that we don't need to add indentation for the case clauses.

src/main/java/Tilo.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public static String readUserInput(Scanner in) {
+        return in.nextLine();
+    }
+
+    public static void addTask(Task[] tasks, int taskCount, String line) {
+        System.out.println(INDENT + "added: " + line);
+        tasks[taskCount] = new Task(line);
+    }
+
```
ycdaniel326: I like how you write separate methods here to make your main function shorter and readable.

src/main/java/Tilo.java
```diff
@@ -3,96 +3,120 @@
...
     }
 
-    public static void addTask(String description) {
-        Task newToDo = new Task(description);
-        tasks.add(newToDo);
+    private void addTask(String description) {
+        Task newTask = new Task(description);
+        tasks.add(newTask);
+        printTaskAdded(newTask);
     }
```
seahsongli: I like the encapsulation implementation here. Should everything here be private?

src/main/java/Tilo.java
```diff
@@ -3,96 +3,120 @@
...
+
+        String description = inputLine.substring(0, fromIndex).trim();
+        String from = inputLine.substring(fromIndex + fromPrefix.length(), toIndex).trim();
+        String to = inputLine.substring(toIndex + toPrefix.length()).trim();
+
+        if (description.isEmpty() || from.isEmpty() || to.isEmpty()) {
+            System.out.println(INDENT + "Description, start time, and end time are all required.");
+            return;
+        }
+
```
seahsongli: I like the idea of error check here. Should the error check be in an enum instead of just a numerical value?

text-ui-test/EXPECTED.TXT
seahsongli, line 1: I like the update of UI Testing case

src/main/java/Tilo.java
```diff
@@ -0,0 +1,220 @@
...
+public class Tilo {
+    private static final int TASK_LIST_SIZE = 100;
+    private static final String INDENT = "\t";
+    private static final String BORDER = "----------------------------------------";
+    private static final String LOGO = INDENT + "___________.___.____     ________\n"
+            + INDENT + "\\__    ___/|   |    |   \\_____  \\\n"
+            + INDENT + "   |    |  |   |    |    /   |   \\\n"
+            + INDENT + "   |    |  |   |    |___/    |    \\\n"
+            + INDENT + "   |____|  |___|_______ \\_______  /\n"
+            + INDENT + "                       \\/       \\/";
```
adoorknob: you could wrap this in """ instead to keep the white spaces

src/main/java/Tilo.java
```diff
@@ -0,0 +1,220 @@
...
+            return true;
+        }
+        if (words[1].trim().isEmpty()) {
+            System.out.println(INDENT + "Invalid input.");
+            return true;
+        }
+        return false;
+    }
+
+    private void printTaskAdded(Task task) {
```
adoorknob: good job having a separate method to print messages! could consider making a whole separate class as well

src/main/java/Tilo.java
```diff
@@ -0,0 +1,220 @@
...
+
+    private void markTask(Task task) {
+        task.markAsDone();
+        System.out.println(INDENT + "Nice! I've marked this task as done:");
+        System.out.println(INDENT + "  " + task);
+    }
+
+    private void unmarkTask(Task task) {
+        task.markAsNotDone();
+        System.out.println(INDENT + "OK, I've marked this task as not done yet:");
```
adoorknob: could avoid using magic strings, would be easier to edit down the road as well

src/main/java/Tilo.java
```diff
@@ -0,0 +1,220 @@
...
+            break;
+        case "unmark":
+            handleMarkCommand(words,false);
+            break;
+        case "bye":
+            isRunning = false;
+            break;
+        default:
+            addTask(userInput);
+            break;
```
adoorknob: i like this switch case, very readable

src/main/java/Tilo.java
```diff
@@ -0,0 +1,220 @@
...
+            System.out.println(INDENT + "Description, start time, and end time are all required.");
+            return;
+        }
+
+        Task newEvent = new Event(description, from, to);
+        tasks.add(newEvent);
+        printTaskAdded(newEvent);
+    }
+
+    private boolean isInputValid(String[] words) {
```
adoorknob: this method is named `isInputValid` but returns true if input is invalid

adoorknob (commented):
🎊 could consider having separate classes for printing messages and creating task types, otherwise good code quality overall! LGTM

PR #55, opened 2025-09-01
[Tan Jun Yi] iP

Tanjy55:


src/main/java/Tank.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    static void markTaskDone(ArrayList<Task> list, String line) {
+        String[] parts = line.split(" ");
+        int ArrayIndex = Integer.parseInt(parts[1]) - 1;
+        list.get(ArrayIndex).setDone();
+        System.out.println("____________________________________________________________");
+        System.out.println("\t Good job! I've marked this task done:");
+        System.out.println("\t\t" + list.get(ArrayIndex).getStatusIcon() + " " + list.get(ArrayIndex).description);
+        System.out.println("____________________________________________________________");
```
kevinlokewy: Do you think it's possible to reduce all these print lines into a method?

src/main/java/Tank.java
```diff
@@ -0,0 +1,77 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+
+public class Tank {
+    static void displayList(ArrayList<Task> list) {
+        int counter = 1;
```
kevinlokewy: I think it makes it easier to read the code if its "taskCounter" instead of just "counter", so the reader is reminded what we're counting

src/main/java/Tank.java
```diff
@@ -0,0 +1,77 @@
...
+        list.get(ArrayIndex).setDone();
+        System.out.println("____________________________________________________________");
+        System.out.println("\t Good job! I've marked this task done:");
+        System.out.println("\t\t" + list.get(ArrayIndex).getStatusIcon() + " " + list.get(ArrayIndex).description);
+        System.out.println("____________________________________________________________");
+    }
+
+    static void markTaskNotDone(ArrayList<Task> list, String line) {
+        String[] parts = line.split(" ");
+        int ArrayIndex = Integer.parseInt(parts[1]) - 1;
```
kevinlokewy: Retrieving the array index like this is kind of confusing at first glance, if we can use a method it might be more readable like getArrayIndex()

src/main/java/Tank.java
```diff
@@ -0,0 +1,77 @@
...
+                markTaskNotDone(listOfTasks, line);
+                continue;
+            }
+
+            if (line.contains(("mark"))) {
+                markTaskDone(listOfTasks, line);
+                continue;
+            }
+
+            if (!line.equals("bye")) {
```
kevinlokewy: I think its better to do 'if-else' instead of 'ifs' because the programme will check through all the 'ifs', but for 'if-else' the programme will exit the loop once it meets one of the conditions.

kevinlokewy: I think its better to do 'if-else' instead of 'ifs' because the programme will check through all the 'ifs', but for 'if-else' the programme will exit the loop once it meets one of the conditions.

src/main/java/Tank.java
```diff
@@ -0,0 +1,77 @@
...
+        System.out.println("What shall we talk about today?\n");
+        System.out.println("____________________________________________________________\n");
+
+        do {
+            line = in.nextLine();
+            if (line.equals(("list"))) {
+                displayList(listOfTasks);
+                continue;
+            }
+
```
mazh25: I have same idea. Maybe it is a better idea to use if-else rather than "if" only.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public void setDone() {
+        isDone = true;
+    }
+
+    public void setNotDone() {
+        isDone = false;
+    }
+
+    public String getStatusIcon() {
```
mazh25: I like this coding style. Using get- and set- to name a method makes code readable.

src/main/java/Tank.java
```diff
@@ -0,0 +1,77 @@
...
+public class Tank {
+    static void displayList(ArrayList<Task> list) {
+        int counter = 1;
+        System.out.println("\tHere are the tasks in your list:");
+        for (Task item : list) {
+            System.out.println("\t" + counter + "." + item.getStatusIcon() + " " + item.description);
+            counter++;
+        }
+        System.out.println("____________________________________________________________");
```
mazh25: Maybe you can find a better way to tackle the print lines.

src/main/java/Tank.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    static void markTaskDone(ArrayList<Task> list, String line) {
+        String[] parts = line.split(" ");
+        int ArrayIndex = Integer.parseInt(parts[1]) - 1;
```
Joannaj00: Variable name should be camel case. (ArrayIndex -> arrayIndex)

mazh25 (commented):
Overall, the coding standards are excellent. I had to look quite carefully to find even minor issues, which speaks to the high quality and attention to detail throughout the code.

src/main/java/Tank.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public static void main(String[] args) {
+
+        String line = "";
+        Scanner in = new Scanner(System.in);
+        ArrayList<Task> listOfTasks = new ArrayList<>();
+        int numberOfTasks = 0;
+
+        System.out.println("____________________________________________________________");
```
Joannaj00: "____________________________________________________________" could be a constant instead of printing every time!!

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public void setDone() {
```
Joannaj00: good job making the method names verbs and camelCase!!

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
Joannaj00: good job making the boolean name sound like a boolean!!

src/main/java/Tank.java
```diff
@@ -0,0 +1,362 @@
...
+
+        try {
+            checkIfStringContains(line, "/by");
+        } catch (Exception e) {
+            System.out.println(e.getMessage() + "/by");
+            printDottedLines();
+            return true;
+        }
+
+        String[] message = processDeadlineInput(line);
```
LuoYu-uwu: Perhaps consider using plural noun instead of message if it always contains multiple parts

src/main/java/Tank.java
```diff
@@ -0,0 +1,362 @@
...
+        }
+    }
+
+    static void checkIfStringContains(String string, String string2) throws TankException {
+        if (!string.contains(string2)) {
+            throw new TankException("Hmmm something went wrong, check if you forgot to include: ");
+        }
+    }
+
+    static boolean checkValidDeadline(String line) {
```
LuoYu-uwu: Perhaps can rename the function to sound more like booleans instead of verb

LuoYu-uwu (commented):
Overall, clean and well structured codes that follows SLAP well. good job!

PR #57, opened 2025-09-01
[Lee Yi Sheng] iP

Lee-YiSheng:


src/main/java/Pep.java
```diff
@@ -1,11 +1,23 @@
+import javax.swing.*;
+import java.util.Scanner;
 public class Pep {
     public static void main(String[] args) {
         String chatbotName = "Pep";
-        System.out.println("____________________________________________________________");
+        String line = "____________________________________________________________";
+        System.out.println(line);
```
sandrej09: The underscores are cool, but maybe you don’t need to repeat them so many times. Could be easier if it was just one variable you reuse.

src/main/java/Pep.java
```diff
@@ -1,11 +1,23 @@
...
         System.out.println(" Hello! I'm " + chatbotName);
         System.out.println(" I love gyming");
-        System.out.println("____________________________________________________________");
+        System.out.println(line);
+        Scanner scanner = new Scanner(System.in);  // Create a Scanner object
+        String userInput =  scanner.nextLine();
+        while (!(userInput.equals("bye"))) {
```
sandrej09: The first nextLine before the loop feels a bit unnecessary. Would it be clearer to use a do { ...}   while (...) loop instead, so you don’t duplicate nextLine() and the program shows something first before reading again?

src/main/java/Pep.java
```diff
@@ -29,14 +29,18 @@ public void run() {
...
+            } else if (userInput.startsWith("todo ")) {
+                String description = userInput.substring(5);
+                Task task = new Todo(description);
+                taskList.addTask(task);
+                ui.showAdded(String.valueOf(task), taskList.getCount());
             } else if (userInput.startsWith("deadline ")) {
-                int index = parseIndex(userInput);
-                if (index != -1 && taskList.unmarkTask(index)) {
-                    ui.showMarked(taskList.getTask(index));
+                String[] parts = userInput.substring(9).split(" /by ", 2);
```
sandrej09: What happens if the user forgets to type /by for a deadline? Maybe you could add some feedback message for that case.

sandrej09: Variable naming like parts could be more descriptive  --  (descriptionAndDate), might make it easier for others to understand quickly.

src/main/java/Ui.java
```diff
@@ -0,0 +1,57 @@
+public class Ui {
```
ycdaniel326, line 13: I like the way that you created a Ui Class for displaying. It makes the code of the main method much shorter and readable.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,40 @@
...
+        this.count = 0;
+    }
+
+    public void addTask(Task task) {
+        if (count < tasks.length) {
+            tasks[count] = task;
+            count++;
+        }
+    }
+
```
ycdaniel326: It is great that you added the checking conditions for each command. It can ensure the program runs smoothly. 
Perhaps you can add some lines to indicate that the action is not feasible, e.g. "TaskList is full!"

src/main/java/Pep.java
```diff
@@ -0,0 +1,40 @@
...
+        ui.showWelcome(chatbotName);
+
+        boolean isExit = false;
+        while (!isExit) {
+            String userInput = scanner.nextLine();
+            try {
+                Command command = Parser.parse(userInput);
+                command.execute(taskList, ui);
+                isExit = command.isExit();
+            } catch (Exception e) {
```
roshnidaksha: Catching `Exception` is too generic. Consider adding some custom/ specific exceptions.

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public void addTask(Task task) {
+        if (count < tasks.length) {
+            tasks[count] = task;
+            count++;
+        }
+    }
+
+    public boolean markTask(int index) {
```
roshnidaksha: Perhaps using a better prefix would make this method sound like Boolean. The same applies to method `unmarkTask` as well.

src/main/java/Parser.java
```diff
@@ -0,0 +1,27 @@
+import commands.*;
+
+public class Parser {
```
roshnidaksha, line 17: `CommandParser` could be a better name

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,66 @@
...
+        printLine();
+    }
+
+    public void showAdded(String task,  int totalTasks) {
+        printLine();
+        System.out.println(" added: " + task);
+        System.out.println(" Now you have " + totalTasks + " tasks in the list.");
+        printLine();
+
+
```
roshnidaksha: Avoid leaving unnecessary blank lines.

roshnidaksha (commented):
Overall, well written code. 
There are some minor things which I have mentioned in the comments.

src/main/java/Parser.java
```diff
@@ -0,0 +1,27 @@
+import commands.*;
```
roshnidaksha: Imported classes should always be listed explicitly. Avoid using wild card import statements.

PR #58, opened 2025-09-01
[abefqy] iP

abefqy:


src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    public void markAsDone() {
```
BTslayer761: naming is good(Easy to identify what the function does)

src/main/java/Darren.java
```diff
@@ -0,0 +1,77 @@
...
+        //greeting
+        System.out.println(line);
+        System.out.println("Hello! I'm " + logo);
+        System.out.println("How can I help you?");
+        System.out.println(line);
+
+        while (true) {
+            String input = scanner.nextLine();
+            //exit feature
+            if (input.equals("bye")) {
```
BTslayer761: could make a switch-case for easier readability of each condition

src/main/java/Darren.java
```diff
@@ -0,0 +1,77 @@
...
+        //get integer for mark and unmark
+        String[] words = input.split(" ");
+        return Integer.parseInt(words[1]);
+    }
+
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        int taskCount = 0;
+        Task[] tasks = new Task[100];
```
BTslayer761: good naming of tasks

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
```
BTslayer761: i dont think need this for this.isDone

src/main/java/Darren.java
```diff
@@ -0,0 +1,77 @@
+import java.util.Scanner;
+
+public class Darren {
+
+    public static Integer getInt(String input) {
```
gordonajajar: I like that you factored out the getInt logic into another method

src/main/java/Darren.java
```diff
@@ -0,0 +1,77 @@
...
+                System.out.println(line);
+                break;
+
+
+            } else if (input.equals("list")) {
+                //list feature
+                System.out.println("Here's your to-do list:");
+                System.out.println(line);
+                for (int i = 0; i < taskCount; i++) {
+                    System.out.println( (i+1) + ". " + tasks[i]);
```
gordonajajar: Formatting here with the spaces is a little off; perhaps turning on "Auto-format on save" in IntelliJ Idea might help?

src/main/java/Darren.java
```diff
@@ -0,0 +1,77 @@
...
+                }
+                System.out.println(line);
+
+
+            } else if (input.startsWith("mark")) {
+                //mark feature
+                Integer index = getInt(input);
+                if (index != null && index <= taskCount && index > 0) {
+                    tasks[index - 1].markAsDone();
+                    System.out.println(line);
```
gordonajajar: I noticed the line printing is a little repetitive; should it be factored out and done just once before and after the case checks?

src/main/java/Darren.java
gordonajajar: Any reason why you used if-else instead of switch cases?

src/main/java/Darren.java
```diff
@@ -0,0 +1,77 @@
...
+                for (int i = 0; i < taskCount; i++) {
+                    System.out.println( (i+1) + ". " + tasks[i]);
+                }
+                System.out.println(line);
+
+
+            } else if (input.startsWith("mark")) {
+                //mark feature
+                Integer index = getInt(input);
+                if (index != null && index <= taskCount && index > 0) {
```
gordonajajar: I like the small input validation here

src/main/java/darren/DarrenException.java
```diff
@@ -0,0 +1,7 @@
+package darren;
+
+public class DarrenException extends Exception {
+    public DarrenException(String message) {
+        super(message);
+    }
+}
```
Daviancold, line 7: Could consider using a more descriptive exception name that describes what it is doing to make it more readable

src/main/java/darren/Darren.java
```diff
@@ -0,0 +1,168 @@
...
+            try {
+                String input = scanner.nextLine();
+                if (input.trim().isEmpty()) {
+                    continue;
+                }
+                // get first word of input as command
+                String[] words = input.split(" ");
+                String command = words[0];
+
+                switch (command) {
```
Daviancold: Could consider extracting the logic in each case statement into their own methods to help with readability and maintainability of code

PR #63, opened 2025-09-01
[ycdaniel326] iP

ycdaniel326:


src/main/java/YC.java
```diff
@@ -0,0 +1,69 @@
+import java.util.Scanner;
```
limzerui: leave a line after your import statements before the start of class declaration

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
limzerui, line 8: Could these be private to keep invariants local?

src/main/java/YC.java
```diff
@@ -0,0 +1,69 @@
...
+public class YC {
+    public static void main(String[] args) {
+        String bot_Name = "YC";
```
limzerui: prefer camelCase without underscores (botName)

limzerui: also consider private static final if constant
```suggestion
       private static final String BOT_NAME + "YC";
       ```

**Review by limzerui — COMMENTED** [review]
code is readable.
suggested a few areas of improvement.

File: src/main/java/Task.java [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
```
yentheng0110: All comments in the code can be removed

src/main/java/YC.java
```diff
@@ -0,0 +1,139 @@
+import java.util.Scanner;
+
+public class YC {
+    public static void main(String[] args) {
+        String botName = "YC";
+        Scanner scanner = new Scanner(System.in);
+        String userCommand;
+        Task[] commandList = new Task[100];
```
yentheng0110: 100 here is a magic number, perhaps you can consider abstracting it out as a constant

src/main/java/YC.java
```diff
@@ -0,0 +1,139 @@
...
+
+        }
+    }
+
+    private static void displayWelcomeMessage(String botName) {
+        System.out.println("\t" + "____________________________________________________________");
+        System.out.println("\t" + "Hello! I'm " + botName);
+        System.out.println("\t" + "What can I do for you?");
+        System.out.println("\t" + "____________________________________________________________" + "\n");
+    }
```
yentheng0110: I like how you abstract each function as a method, good job!

src/main/java/YC.java
```diff
@@ -0,0 +1,139 @@
...
+        System.out.println("\t" + "Got it. I've added this task:");
+        System.out.println("\t\t" + commandList[totalCommands].toString());
+        totalCommands++;
+        System.out.println("\t" + "Now you have "+ totalCommands + " tasks in the list.");
+        System.out.println("\t" + "____________________________________________________________" + "\n");
+        return totalCommands;
+    }
+
+    private static int ProcessTodo(String userCommand, Task[] commandList, int totalCommands) {
+        String description = userCommand.substring(5);
```
yentheng0110: The number 5 is a magic number, consider abstracting it out as a constant. The same applies to other methods

yentheng0110 (commented):
In general, your code is clean with proper method abstraction, minor changes were suggested.

PR #77, opened 2025-09-02
[Ong Chong Yao] iP

chongyaoo:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+    private String date;
+
+    public Deadline (String data) {
+        super(data);
+        this.isMarked = false;
+    }
+
+    @Override
+    public String toType() {
+        return "[D]";
```
msc-123456, line 24: I think you can type this like "return "[D]" + super.toString() + " (by: " + by + ")";"?

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,97 @@
+import java.util.*;
+
+public class Grizzly {
+    public static void main(String[] args) {
+        String logo = """
+                Hi, I'm your friendly assistant, Grizzly!
+                What can I do for you today?
+                ____________________________________________________________""";
```
msc-123456: maybe you can add more lines to seperate each action?

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,97 @@
...
+                    System.out.println("Now you have " + index + " tasks in the list");
+                    index++;
+                    break;
+                default:
+                    break;
+            }
+            line = in.nextLine();
+        }
+    }
+}
```
msc-123456: this part is a little complex, I think it's better to use functions to handle different situations

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,97 @@
+import java.util.*;
+
+public class Grizzly {
+    public static void main(String[] args) {
+        String logo = """
+                Hi, I'm your friendly assistant, Grizzly!
```
msc-123456: maybe you can add a logo like the given "Duke"

msc-123456 (commented):
it's good to use separate files to make todo、event and deadline, but the main part is too complex

src/main/java/Grizzly.java
ChangIkJoong, line 96: This part I would suggest to abstract into a class perhaps handling the output, the other the logic. 
Right now it seems a bit overly complex and hard to follow.

ChangIkJoong: This part I would suggest to abstract into a class perhaps handling the output, the other the logic.
Right now it seems a bit overly complex and hard to follow.

src/main/java/Task.java
```diff
@@ -0,0 +1,41 @@
...
+        this.isMarked = marked;
+    }
+
+    public String MarkedBox() {
+        if (isMarked) {
+            return "[X] ";
+        } else {
+            return "[ ] ";
+        }
+    }
```
ChangIkJoong, line 50: I would keep the output of let's say a string seperate and just keep this as a boolean within the class, then another class perhaps handling the string output.

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,97 @@
...
+                        itemToMark = Integer.parseInt(parts[1]);
+                    } catch (NumberFormatException e) {
+                        System.out.println("Not a valid integer.");
+                        break;
+                    }
+                    if (itemToMark < index && itemToMark >= 1) {
+                        storedItems[itemToMark - 1].setMarked(true);
+                        System.out.println("Nice! I've marked this task as done:");
+                        System.out.println(itemToMark + "." + storedItems[itemToMark - 1].toType() + storedItems[itemToMark - 1].MarkedBox() + storedItems[itemToMark - 1].getString());
+                    } else {
```
ChangIkJoong: Feels like this is a bit too nestled, perhaps the if function can be abstracted to it's own method for improved readability?

src/main/java/Task.java
```diff
@@ -0,0 +1,41 @@
+public class Task {
+    protected String data;
```
ChangIkJoong, line 12: Maybe the name data can be a bit too open, doesn't describe too much what kind of data or what it is.

ChangIkJoong (commented):
Overall nice coding, I like the inheritance. I would suggest to abstract some methods and increase readability for certain parts as provided, as some parts had perhaps a bit too deep nesting and overall a bit hard to read some functional aspects as they became quite long.

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,97 @@
...
+        System.out.println(logo);
+
+        Scanner in = new Scanner(System.in);
+        String line;
+        int index = 1;
+        Task[] storedItems = new Task[100];
+        line = in.nextLine();
+
+
+        while (true) {
```
JianHao24, line 36: The code can be refactored by extracting the code in each cases, this would promote better abstraction and support readability.

src/main/java/Task.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public boolean isMarked() {
+        return isMarked;
+    }
+
+    public void setMarked(boolean marked) {
+        this.isMarked = marked;
+    }
+
+    public String MarkedBox() {
```
JianHao24: method MarkedBox() should follow lowerCamelCase, perhaps rename as markedBox() or getMarkedBox() since it returns a value.

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,97 @@
...
+        String logo = """
+                Hi, I'm your friendly assistant, Grizzly!
+                What can I do for you today?
+                ____________________________________________________________""";
+        System.out.println(logo);
+
+        Scanner in = new Scanner(System.in);
+        String line;
+        int index = 1;
+        Task[] storedItems = new Task[100];
```
JianHao24: Can use a constant for the maximum array size instead of an arbitrary number 100 for better readability.  Maybe define a static final MAX_TASKS = 100 at the start.

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,97 @@
...
+                    } else {
+                        System.out.println("Task " + itemToMark + " has not been created!");
+                    }
+                    break;
+                case "unmark":
+                    int itemToUnmark;
+                    try {
+                        itemToUnmark = Integer.parseInt(parts[1]);
+                    } catch (NumberFormatException e) {
+                        System.out.println("Not a valid integer.");
```
JianHao24: Maybe you can define your errors as constants at the start, maybe naming it something like INVALID_INT_ERROR. This promotes readability of code and also reduces redundancy should the error message be reused. The constant also helps standardise error message in case one accidentally varies it.

src/main/java/ToDo.java
```diff
@@ -0,0 +1,12 @@
+public class ToDo extends Task {
+
+    public ToDo (String data) {
+        super(data);
+        this.isMarked = false;
+    }
+
+    @Override
+    public String toType() {
```
JianHao24, line 19: You can overload the toString function to print the boxes which mark and represent class type when printing the instance instead of using a seperate toType function.

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,150 @@
...
+        int index = 1;
+        Task[] storedItems = new Task[100];
+        line = in.nextLine();
+
+
+        while (true) {
+            String[] parts = line.split("\\s+", 2);
+            String action = parts[0].toLowerCase();
+            switch (action) {
+                case "bye":
```
syCHEN1645: Maybe consider refactor magic literals with final strings?

src/main/java/Grizzly.java
```diff
@@ -0,0 +1,150 @@
...
+                            Bye. Hope to see you again soon!
+                            """;
+                    System.out.println(logo);
+                    return;
+                case "mark":
+                    int itemToMark;
+                    try {
+                        itemToMark = Integer.parseInt(parts[1]);
+                    } catch (NumberFormatException e) {
+                        System.out.println("Not a valid integer.");
```
syCHEN1645: This message seems to be used in other cases too, maybe define a constant variable for it?

syCHEN1645 (commented):
Good effort! Your code looks clear, just a bit too bulky. It would be graeat if you can break it down to smaller methods!

PR #79, opened 2025-09-02
Xylon Chan Yi Chong

xylonc:


src/main/java/Victoria.java
```diff
@@ -19,32 +19,78 @@ public static void echo(String[] args){
     }
 
     public static void addList(String[] args){
-        String[] list = new String[100];
+        Task[] list = new Task[100];
         int count =0;
```
XuHh03: is it better to add a space?

src/main/java/Victoria.java
```diff
@@ -19,32 +19,78 @@ public static void echo(String[] args){
...
         while(true) {
-            Scanner scanner = new Scanner(System.in);
             task = scanner.nextLine();
             if (task.equals("list")) {
+                System.out.println("-".repeat(30));
+                System.out.println("Here are your tasks:");
                 for (int i = 0; i < count; i++) {
                     System.out.println((i + 1) + ". " + list[i]);
                 }
                 System.out.println("\n" + "-".repeat(30));
```
XuHh03: I like the use of repeat instead of typing out the whole String, but would you prefer creating a method that does this so that you don't have to repeat it throughout the file

src/main/java/Victoria.java
```diff
@@ -19,32 +19,78 @@ public static void echo(String[] args){
...
+        protected String description;
+        protected boolean isDone;
+
+        public Task(String description) {
+            this.description = description;
+            this.isDone = false;
+        }
+
+        public String getStatusIcon() {
+            return (isDone ? "X" : " "); // mark done task with X
```
XuHh03: Will it be better to add the comment on top of this line 
```suggestion
// mark done task with X
return (isDone ? "X" : " ");
```

src/main/java/Victoria.java
```diff
@@ -19,32 +19,78 @@ public static void echo(String[] args){
...
+                System.out.println("-".repeat(30));
+                System.out.println("Here are your tasks:");
                 for (int i = 0; i < count; i++) {
                     System.out.println((i + 1) + ". " + list[i]);
                 }
                 System.out.println("\n" + "-".repeat(30));
-            } else if(task.equals("bye")) {
+            }
+            else if(task.startsWith("mark")) {
+                int taskNum = Integer.parseInt(task.split(" ")[1]) - 1;
```
XuHh03: Would it be better to write in full as taskNumber?

src/main/java/Victoria.java
```diff
@@ -19,32 +19,78 @@ public static void echo(String[] args){
...
                 System.out.println("\n" + "-".repeat(30));
-                list[count] = task;
+                list[count] = new Task(task);
                 count++;
                 System.out.println("added:" + task + "\n");
                 System.out.println("-".repeat(30) + "\n");
             }
         }
     }
+    public static class Task {
```
XuHh03: Will it be clearer to declare this in a new file called task.java?

src/main/java/Victoria.java
```diff
@@ -0,0 +1,114 @@
...
+            System.out.println("-".repeat(30) + "\n");
+            echo(args);
+        }
+    }
+
+    public static void addList(String[] args){
+        Task[] list = new Task[100];
+        int count =0;
+        String task;
+        Scanner scanner = new Scanner(System.in);
```
ZT712002: Scanner should be a constant (FINAL) and be in all cases. See stack overflow for reasons why (security reasons)

src/main/java/Victoria.java
```diff
@@ -0,0 +1,114 @@
...
+
+
+        public String toString() {
+            return "[" + getStatusIcon() + "] " + description;
+        }
+    }
+
+
+    public static void main(String[] args) {
+        String logo =
```
ZT712002, line 29: I like the added ascii art as an added flavor text. Nice.

src/main/java/Victoria.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public static void addList(String[] args){
+        Task[] list = new Task[100];
+        int count =0;
+        String task;
+        Scanner scanner = new Scanner(System.in);
+        while(true) {
+            task = scanner.nextLine();
+            if (task.equals("list")) {
```
ZT712002: Maybe a switch statement here would be better for readability

src/main/java/Victoria.java
```diff
@@ -0,0 +1,114 @@
+import java.util.Scanner;
+
+public class Victoria {
+    public static void echo(String[] args){
+        String answer;
+        Scanner reply = new Scanner(System.in);
+        answer = reply.nextLine();
+        if(answer.equals("bye")) {
+            System.out.println("-".repeat(30));
```
ZT712002: I see many of the same "System.out.println("-".repeat(30));", maybe creating a function for it can help reduce redundancy

ZT712002 (approved):
Overall, there are no issues with the code in question here. I found your code easy to read with some caveats (i.e many repeated prints "-"). You may also consider a "Passer class" to handle user inputs and the scannling of them.

src/main/java/AddList.java
```diff
@@ -0,0 +1,81 @@
+import java.util.Scanner;
+public class AddList {
+    public static void main(String[] args) throws Exception {
+        Task[] list = new Task[100];
```
namsengi11: Could consider factoring out the constant 100 as MAX_TASKS_LENGTH

src/main/java/AddList.java
```diff
@@ -0,0 +1,81 @@
...
+        int count = 0;
+        String task;
+        Scanner scanner = new Scanner(System.in);
+        while (true) {
+            task = scanner.nextLine();
+            if (task.isEmpty()) throw new EmptyStringException();
```
namsengi11: Please avoid using single line if statements

src/main/java/AddList.java
```diff
@@ -0,0 +1,81 @@
...
+                System.out.println(" " + list[taskNum]);
+                System.out.println("-".repeat(30));
+            } else if (task.startsWith("unmark")) { //unmark
+                int taskNum = Integer.parseInt(task.split(" ")[1]) - 1;
+                list[taskNum].markUndone();
+                System.out.println("-".repeat(30));
+                System.out.println("Shag , I have marked it as undone!");
+                System.out.println(" " + list[taskNum]);
+                System.out.println("-".repeat(30));
+            }else if(task.startsWith("deadline")) {//adding deadline
```
namsengi11: Missed space between bracket and if statement.

src/main/java/AddList.java
```diff
@@ -0,0 +1,81 @@
...
+                System.out.println(list[count]);
+                System.out.println("now you got: " + (count + 1) + " tasks");
+                System.out.println(" ");
+                System.out.println("-".repeat(30));
+                count++;
+            }else if(task.startsWith("event")) {
+                int indexFrom = task.indexOf("from");
+                int indexTo = task.indexOf("to");
+                int indexSlash = task.indexOf("/");
+                String From = task.substring(indexFrom + 4 ,indexTo - 1 ); //takes timing between "from" and "to"
```
namsengi11: Good job leaving comment for what the offsets in index represent.

src/main/java/AddList.java
```diff
@@ -0,0 +1,81 @@
...
+                System.out.println(" ");
+                System.out.println("-".repeat(30));
+                count++;
+            } else if (task.equals("bye")) { //exiting the code
+                System.out.println("\n" + "-".repeat(30));
+                System.out.println("Bye.Hope to see you again soon!\n");
+                System.out.println("-".repeat(30));
+                break;
+            } else if(task.startsWith("todo")){ //adding task with no deadline
+                //System.out.println("\n" + "-".repeat(30));
```
namsengi11: Please remove commented codes that are not being used anymore

PR #93, opened 2025-09-02
[andrewcai8] iP

andrewcai8:


src/main/java/Grace.java
```diff
@@ -1,10 +1,29 @@
+import java.util.Scanner;
+
+
 public class Grace {
     public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
```
fahadmohaideen: I feel that for repeated strings like the partition line, you could have saved it in a variable

src/main/java/Grace.java
```diff
@@ -1,10 +1,29 @@
...
         System.out.println("____________________________________________________________");
-        System.out.println("Bye. Hope to see you again soon!");
-        System.out.println("____________________________________________________________");
+
+
+
+        while (true){
+            String input = sc.nextLine();
+            if (input.equals("bye")){
+                System.out.println("____________________________________________________________");
```
fahadmohaideen: It is good that you organised this chunk of code such that each line you are printing is on a different line of code, instead of multiple string concatenations. It makes it easier to visualise the output.

src/main/java/Grace.java
```diff
@@ -19,6 +19,12 @@ public static void main(String[] args) {
                 System.out.println("Bye. Hope to see you again soon!");
                 System.out.println("____________________________________________________________");
                 break;
+            } else if (input.equals("list")){
```
fahadmohaideen: I think it is good that you placed the desirable command path ("happy path") first inside the if statement, while placing the exceptions/edge cases inside the else block

src/main/java/Grace.java
```diff
@@ -1,5 +1,23 @@
...
+class Task{
+    private String description;
+    private boolean isDone;
+
+    public Task(String description){
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public void mark(){
```
fahadmohaideen: I think it is good that each of your class methods only does one specific task. Like for eg, mark() only marks the task as done, and unmark() only marks it as not done (false)

src/main/java/Grace.java
```diff
@@ -43,8 +45,28 @@ public static void main(String[] args) {
...
+                int index = Integer.parseInt(input.split(" ")[1]) - 1;
+                if (index < taskCount && index >= 0){
+                    tasks[index].mark();
+                    System.out.println("____________________________________________________________");
+                    System.out.println(" Nice! I've marked this task as done:");
+                    System.out.println(" " + tasks[index]);
+                    System.out.println("____________________________________________________________");
+                }
+            } else if (input.startsWith("unmark ")){
+                int index = Integer.parseInt(input.split(" ")[1]) - 1;
```
fahadmohaideen: I felt that you could simplify expressions like this. Perhaps you could save the output of the input.split() into a variable before passing it into the parseInt() method

src/main/java/Grace.java
```diff
@@ -0,0 +1,55 @@
...
+        int taskCount = 0;
+
+        System.out.println("____________________________________________________________");
+        System.out.println("Hello! I'm Grace!!");
+        System.out.println("What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        while (true) {
+            String input = sc.nextLine();
+            if (input.equals("bye")) {
```
fookang: Methods can be abstracted out for better readability

src/main/java/Grace.java
```diff
@@ -0,0 +1,55 @@
...
+                System.out.println("Bye. Hope to see you again soon!");
+                System.out.println("____________________________________________________________");
+                break;
+            } else if (input.equals("list")) {
+                System.out.println("____________________________________________________________");
+                for (int i = 0; i < taskCount; i++) {
+                    System.out.println(" " + (i + 1) + ". " + tasks[i]);
+                }
+                System.out.println("____________________________________________________________");
+            } else if (input.startsWith("mark ")) {
```
fookang: Nested if, switch statement will make the code more readable

src/main/java/Grace.java
```diff
@@ -0,0 +1,55 @@
+import java.util.Scanner;
+
+public class Grace {
+    public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
+        Task[] tasks = new Task[100];
```
fookang: Best to avoid magic number

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
+public class Task {
+    private final String description;
+    private boolean isDone;
```
fookang: Good naming convention in general!

src/main/java/Grace.java
danielkwan2004: Try to use a constant e.g. MAX_TASKS = 100

src/main/java/Grace.java
```diff
@@ -0,0 +1,55 @@
...
+        int taskCount = 0;
+
+        System.out.println("____________________________________________________________");
```
danielkwan2004: This string repeats many times. Can consider using a constant HLINE = "___________________________________" and then print via System.out.println(HLINE);

src/main/java/Grace.java
```diff
@@ -0,0 +1,55 @@
...
+                System.out.println("____________________________________________________________");
+                break;
+            } else if (input.equals("list")) {
+                System.out.println("____________________________________________________________");
+                for (int i = 0; i < taskCount; i++) {
+                    System.out.println(" " + (i + 1) + ". " + tasks[i]);
+                }
+                System.out.println("____________________________________________________________");
+            } else if (input.startsWith("mark ")) {
+                int index = Integer.parseInt(input.split(" ")[1]) - 1;
```
danielkwan2004: Split once, reuse the array and validate length before accessing [1].

src/main/java/Task.java
danielkwan2004: Boolean naming is good. reads like English per standard

src/main/java/Grace.java
```diff
@@ -0,0 +1,55 @@
...
+        Scanner sc = new Scanner(System.in);
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+
+        System.out.println("____________________________________________________________");
+        System.out.println("Hello! I'm Grace!!");
+        System.out.println("What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        while (true) {
```
danielkwan2004: Braces and 4-space indentation look correct.

danielkwan2004 (commented):
Overall: good naming, braces, imports, array style and boolean naming.
Things to improve: constants for magic values and repeated strings

src/main/java/grace/Grace.java
```diff
@@ -0,0 +1,155 @@
...
+                throw new GraceException("A deadline need both a description and a /by time.");
+            }
+            task = new Deadline(parts[0].trim(), parts[1].trim());
+        } else if (input.startsWith("event")) {
+            String[] parts = input.substring(5).split("/from |/to ");
+            if (parts.length < 3 || parts[0].trim().isEmpty() || parts[1].trim().isEmpty() || parts[2].trim().isEmpty()) {
+                throw new GraceException("An event needs a description, /from time, and a /to time.");
+            }
+            task = new Event(parts[0].trim(), parts[1].trim(), parts[2]);
+        }
```
Daviancold: Could consider using a switch case to improve readability of the code

src/main/java/grace/Grace.java
```diff
@@ -0,0 +1,155 @@
...
+        try {
+            int index = Integer.parseInt(input.split(" ")[1]) - 1;
+            if (index >= 0 && index < taskCount) {
+                return index;
+            }
+        } catch (Exception e) {
+            return -1;
+        }
+        return -1;
+    }
```
Daviancold: Alternatively, you could consider catching an IndexOutOfBound exception

src/main/java/grace/Grace.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+        sc.close();
+    }
+
+    private static void greet() {
+        printLine();
+        printMessage("Hello! I'm grace.Grace!!");
+        printMessage("What can I do for you?");
+        printLine();
+    }
```
Daviancold: Can consider using a method for printing to the command line, where you pass in the message and it appends the line for you, to reduce repeated line printing

PR #97, opened 2025-09-02
[Jewel Jace Lim] iP

Jeweljace:


src/main/java/Helio.java
```diff
@@ -0,0 +1,151 @@
...
+                System.out.println(" Got it. I've added this task:");
+                System.out.println("   " + userList[count - 1]);
+                System.out.println(" Now you have " + count + " tasks in the list.");
+                System.out.println("____________________________________________________________");
+            } else if (input.startsWith("event ")) {
+                String[] sections = input.substring(6).split(" /from | /to ");
+                if (sections[0].isEmpty() || sections[1].isEmpty() || sections[2].isEmpty()) {
+                    System.out.println("____________________________________________________________");
+                    System.out.println(" Meow! Please fill in a description and date/time of start and end! =.=");
+                    System.out.println("____________________________________________________________");
```
gbinw128: Maybe it would be better to replace all the if-else case with different functions or a function with parameters?

src/main/java/Helio.java
```diff
@@ -0,0 +1,151 @@
+import java.util.Scanner;
+
+public class Helio {
+    public static void main(String[] args) {
+        Task[] userList = new Task[100];
+        int count = 0;
+        String logo = "        ╱|\n" +
+                "       (˚ˎ 。7\n" +
+                "        |、˜〵\n" +
+                "       じしˍ,)ノ\n";
```
gbinw128: cute cat!

src/main/java/Helio.java
```diff
@@ -0,0 +1,151 @@
...
+                    continue;
+                }
+                userList[count] = new Todo(description);
+                count++;
+                System.out.println("____________________________________________________________");
+                System.out.println(" Got it. I've added this task:");
+                System.out.println("   " + userList[count - 1]);
+                System.out.println(" Now you have " + count + " tasks in the list.");
+                System.out.println("____________________________________________________________");
+            } else if (input.startsWith("deadline ")) {
```
gbinw128: All "else" is in line with the close bracket, nice!

src/main/java/helio/task/Task.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    public void markAsNotDone() {
+        isDone = false;
+    }
+
```
gbinw128, line 44: Nicely named functions

Amanda-HUANG88:
Very thorough consideration！Maybe it is better to do some refactoring to reduce redudant codes and extract methods.

src/main/java/Helio.java
```diff
@@ -0,0 +1,151 @@
+import java.util.Scanner;
+
+public class Helio {
+    public static void main(String[] args) {
+        Task[] userList = new Task[100];
+        int count = 0;
+        String logo = "        ╱|\n" +
```
AWeiBin: Since the logo is a constant, I suggest that it can be a final class level constant

src/main/java/Helio.java
```diff
@@ -0,0 +1,151 @@
...
+                "        |、˜〵\n" +
+                "       じしˍ,)ノ\n";
+        System.out.println("____________________________________________________________\n" +
+                " Hello... I'm Helio o.o\n" +
+                logo +
+                " What can I do for you? ._.\n" +
+                "____________________________________________________________");
+
+        Scanner in = new Scanner(System.in);
+        while (in.hasNextLine()) {
```
gbinw128: Entire loop looks like an arrowhead code

src/main/java/Helio.java
```diff
@@ -0,0 +1,151 @@
...
+            if (input.isEmpty()) {
+                continue;
+            }
+            System.out.println(" Command entered: " + input);
+            if (input.equals("bye")) {
+                System.out.println("____________________________________________________________\n" +
+                        " Bye. Hope to see you again soon!\n" +
+                        "____________________________________________________________");
+                break;
+            } else if (input.equals("help")) {
```
AWeiBin: Good Job on making a help command

gbinw128 (commented):
Overall, nice!

src/main/java/Helio.java
```diff
@@ -0,0 +1,151 @@
+import java.util.Scanner;
+
+public class Helio {
+    public static void main(String[] args) {
+        Task[] userList = new Task[100];
```
AWeiBin: Perhaps the variable name can be more specific.

src/main/java/Helio.java
```diff
@@ -0,0 +1,151 @@
...
+        int count = 0;
+        String logo = "        ╱|\n" +
+                "       (˚ˎ 。7\n" +
+                "        |、˜〵\n" +
+                "       じしˍ,)ノ\n";
+        System.out.println("____________________________________________________________\n" +
```
AWeiBin: Consider making a function to print the borders

src/main/java/helio/ui/Handle.java
```diff
@@ -0,0 +1,42 @@
+public final class Handle {
+    private Handle() {
+    }
```
roshnidaksha: If a constructor is not required for this class, you can remove these lines.

src/main/java/helio/ui/Handle.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private static final String LINE = "____________________________________________________________";
+
+    public static void invalidCommand() {
```
roshnidaksha: Method names should be a verb. Can you modify it to sound like a verb?

src/main/java/Helio.java
```diff
@@ -0,0 +1,112 @@
...
+            } else if (input.equals("list")) {
+                if (count == 0) {
+                    Message.emptyList();
+                } else {
+                    Message.listHeader();
+                    for (int i = 0; i < count; i++) {
+                        System.out.println((i + 1) + ". " + userList[i]);
+                    }
+                    Message.listFooter();
+                }
```
roshnidaksha: Implementation details of these features can be extracted out to a new method to improve SLAP.
The same comment goes for your other features as well.

src/main/java/helio/Helio.java
```diff
@@ -0,0 +1,112 @@
...
+                    Message.emptyList();
+                } else {
+                    Message.listHeader();
+                    for (int i = 0; i < count; i++) {
+                        System.out.println((i + 1) + ". " + userList[i]);
+                    }
+                    Message.listFooter();
+                }
+            } else if (input.startsWith("mark ")) {
+                String check = input.substring(5).trim();
```
roshnidaksha: Consider creating a new method for low-level string operations too.

src/main/java/helio/ui/Handle.java
```diff
@@ -0,0 +1,42 @@
+public final class Handle {
```
roshnidaksha: Does this need to be inside `Message.java`? Both the classes only print messages.

roshnidaksha (commented):
The code is neat generally. 
Some methods can be named better.
You can consider improving SLAP of overall code as well.

PR #101, opened 2025-09-03
[Daniel Kwan] iP

danielkwan2004:


src/main/java/Task.java
```diff
@@ -0,0 +1,19 @@
+// Add the ability to store whatever text entered by the user and display them back to the user when requested.
+public class Task {
+    
+
```
yikjin: Too many blank lines (has 2, should be 1 or none)


```suggestion

```

src/main/java/ChatTPG.java
```diff
@@ -0,0 +1,27 @@
+import java.util.*;
```
yikjin: Imported classes should always be added explicitly, see the [respective Java coding standard](https://se-education.org/guides/conventions/java/basic.html#package-and-import-statements)

src/main/java/ChatTPG.java
```diff
@@ -0,0 +1,27 @@
...
+    public static void main(String[] args) {
+        System.out.println("Hello! I'm ChatTPG");
+        System.out.println("What can I do for you?");
+        
+        Scanner scanner = new Scanner(System.in);
+        String userInput = "";
+        
+        Task task = new Task();
+
+        while (!userInput.equals("bye")) {
```
yikjin: Unnecessary check here, since you already checked for this condition a few lines down


```suggestion
        while (true) {
```

src/main/java/Task.java
```diff
@@ -0,0 +1,19 @@
...
+    String[] taskList = new String[100];
+    int taskCount = 0;
```
yikjin: Variable names should be plural


```suggestion
    String[] tasksList = new String[100];
    int tasksCount = 0;
```

src/main/java/Task.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public void addTask(String description){
```
yikjin: Space needed before `{`


```suggestion
    public void addTask(String description) {
```

src/main/java/Task.java
```diff
@@ -0,0 +1,19 @@
...
+        taskList[taskCount] = description;
+        taskCount++;
+        System.out.println("\tadded: " + description);
+    }
+
+    public void listTask(){
```
yikjin: Space needed before `{`, and method name should be plural


```suggestion
    public void listTasks() {
```

yikjin (commented):
An effort to stick to the coding style can be seen, good job! A few changes requested.

src/main/java/chattpg/logic/TaskOrganiser.java
```diff
@@ -0,0 +1,166 @@
...
+    public void printWelcomeMessage() {
+        System.out.println("Welcome to Task Organiser!");
+        System.out.println("You can manage your tasks here.");
+        printAvailableCommands();
+        System.out.println("Please enter your command:");
+    }
+
+
+    public void deleteTask(int taskNumber) throws TaskIndexOutOfBoundsException {
+        if (taskNumber < 1 || taskNumber > taskCount) {
```
namsengi11: Good job with error handling

src/main/java/chattpg/logic/TaskOrganiser.java
```diff
@@ -0,0 +1,166 @@
...
+            }
+            tasks[taskCount] = new Deadline(parts[0].trim(), parts[1].trim());
+        } else if (description.startsWith("event")) {
+            int fromPos = description.indexOf(" /from ");
+            int toPos = description.indexOf(" /to ");
+            if (fromPos == -1 || toPos == -1 || fromPos > toPos) {
+                throw new InvalidCommandException("event format: event <desc> /from <start> /to <end>");
+            }
+            String desc = description.substring(0, fromPos).trim();
+            String start = description.substring(fromPos + 7, toPos).trim();
```
namsengi11: Try to avoid use of magic numbers. At first glance, the reason for indexing of substrings is unclear.

src/main/java/chattpg/logic/TaskOrganiser.java
```diff
@@ -0,0 +1,166 @@
...
+        printWelcomeMessage();
+        
+        while (true) {
+            final String userInput = scanner.nextLine().trim();
+            try {
+                switch (userInput) {
+                    case "help":
+                        printAvailableCommands();
+                        break;
+                    case "mark done":
```
namsengi11: You could refactor the logic for each case in order to make the run loop more readable.

src/main/java/chattpg/logic/TaskOrganiser.java
```diff
@@ -0,0 +1,166 @@
...
+            System.out.println("You have no tasks in your list.");
+            return;
+        }
+        System.out.println("Here are the tasks in your list: ");
+        for (int i = 0; i < taskCount; i++) {
+            System.out.println("\t" + (i + 1) + ". " + tasks[i].toString());
+        }
+    }
+
+    public void taskAdded() {
```
namsengi11: Method could be named to better indicate action responsible, such as announceTaskAdded()

src/main/java/chattpg/model/Deadline.java
```diff
@@ -0,0 +1,17 @@
+package chattpg.model;
+import static chattpg.model.Task.requireNonBlank;
+
+public class Deadline extends Task {
+
+    private final String by;
```
namsengi11, line 9: While I understand why the variable could be named 'by', but it should be more descriptive like 'dueDate'

PR #103, opened 2025-09-03
[Ibrahim Shoukry] iP

Ibrashoukry:


src/main/java/Mark1.java
```diff
@@ -7,11 +8,23 @@ public static void main(String[] args) {
 
         Scanner in = new Scanner(System.in);
         String command = in.nextLine(); // initial request
+        Vector<String> tasks = new Vector<>();
```
patrykmrozek: Why do you choose a vector?
You could perhaps use an ArrayList instead  - faster and safer

src/main/java/Mark1.java
```diff
@@ -7,11 +8,23 @@ public static void main(String[] args) {
...
 
         // do tasks if user does not want to exit
         while (!command.contains("bye")) {
-            System.out.println("\n" + lineBreak + command + "\n" + lineBreak);
+            if (command.equals("list")) {
+                System.out.println(lineBreak);
+                for (String task : tasks) {
```
patrykmrozek: I suggest iterating using an index instead of with the task.
When you print you could consider using (i+1) and tasks[i] instead of tasks.indexOf(task) and task
- more readable, cleaner, shorter

src/main/java/Mark1.java
```diff
@@ -7,11 +8,23 @@ public static void main(String[] args) {
 
         Scanner in = new Scanner(System.in);
         String command = in.nextLine(); // initial request
+        Vector<String> tasks = new Vector<>();
 
         // do tasks if user does not want to exit
         while (!command.contains("bye")) {
```
patrykmrozek: Good job using effective string methods.

src/main/java/Mark1.java
```diff
@@ -0,0 +1,35 @@
...
+    public static void main(String[] args) {
+        String lineBreak = "~~~~~~~~~~~~~~~~~~~~~~\n";
+        System.out.println(lineBreak + "Hello I'm Mark I\n" + "Alssalam Alaikum. I'm here to assist you *_*\n" + lineBreak);
+
+        Scanner in = new Scanner(System.in);
+        String command = in.nextLine(); // initial request
+        Vector<String> tasks = new Vector<>();
+
+        // do tasks if user does not want to exit
+        while (!command.contains("bye")) {
```
ashpasa: It may be beneficial to find alternative termination conditions for the loop: consider the case where the user wants to input the task "say bye to aunt before she goes overseas"

src/main/java/abdo/ui/Abdo.java
```diff
@@ -0,0 +1,88 @@
...
+
+        while (!command.equals("bye")) {
+            String[] parsedCommand = command.split(" ");
+            switch(parsedCommand[0]) {
+            case "list":
+                System.out.println(printer.lineBreak + "Here are the tasks in your list:");
+                for (int i = 0; i < nextTask; i++) {
+                    System.out.println((i + 1) + ". " + tasks[i].toString());
+                }
+                System.out.print(printer.lineBreak);
```
roshnidaksha: Consider moving the implementation details to a new method for better SLAP.

src/main/java/Abdo.java
```diff
@@ -0,0 +1,88 @@
...
+        Task task;
+
+        String unparsedDesc = command.substring(type.length() + 1);
+
+        switch(type) {
+        case "todo":
+            task = new Todo(unparsedDesc);
+            break;
+        case "deadline":
+            parsedDesc = unparsedDesc.split("/by");
```
roshnidaksha: These low-level string operations can be moved to a new method too.

src/main/java/abdo/process/Task.java
```diff
@@ -0,0 +1,22 @@
+public class Task {
```
roshnidaksha: Since Task itself is never directly instantiated and objects are always Todo, Deadline, or Event, it would be better to declare Task as abstract. This makes the design clearer and prevents accidental creation of plain Task objects.

src/main/java/abdo/task/Task.java
```diff
@@ -0,0 +1,22 @@
...
+
+    protected boolean isDone;
+    protected String description;
+
+    public Task(String description) {
+        this.description = description;
+        isDone = false;
+    }
+
+    public String getStatusIcon (boolean isDone) {
```
roshnidaksha, line 13: Space is not necessary between method name and opening bracket for parameters. 
Try maintaining consistency with spacing for all methods too.

roshnidaksha (commented):
Overall, code can be better refactored with SLAP in mind. Otherwise, code is neat.

PR #110, opened 2025-09-04
[sivanshno] ip

sivanshno:


src/main/java/Agus.java
```diff
@@ -0,0 +1,92 @@
+import java.util.Scanner;
+public class Agus {
+
+    public static final String [] VALID_COMMANDS = { "bye", "list", "mark", "unmark", "todo", "deadline", "event"};
```
yentheng0110: You can perhaps remove the spacing after {

src/main/java/Agus.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+    }
+
+    public static int extractNumber(String userInput){
+        String [] words = userInput.split(" ");
+        return Integer.parseInt(words[1]);
+    }
+
+    // helper function to determine whether the user input is a command
```
yentheng0110: All comments in the code can be removed

src/main/java/Agus.java
```diff
@@ -0,0 +1,92 @@
...
+    public static boolean isCommand(String userInput){
```
yentheng0110: I like how you name the variable such that it sounds like a boolean variable (isCommand)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,21 @@
+public class Deadline extends Task{
+
+    private String deadline;
+
+    public Deadline(String userInput){
+        super(userInput);
+        int dateStartIndex = userInput.indexOf("/") + 4;
+        description = userInput.substring(0,dateStartIndex - 5);
```
yentheng0110, line 23: 4 and 5 here are magic numbers, perhaps you could consider abstracting them out as constants
Same goes to your other classes

yentheng0110 (commented):
In general, the code is clean, minor changes were suggested

